PR #9, opened 2021-03-16
Ch83715/replace event stream

piercetrey-figure:


p8e-api/src/main/kotlin/io/provenance/engine/stream/ScopeStream.kt
```diff
@@ -106,18 +87,15 @@ class ScopeStream(
...
         val lastHeight = record?.lastBlockHeight
             ?: transaction { EventStreamRecord.insert(eventStreamId, epochHeight) }.lastBlockHeight
-        val request = EventProtos.EventStreamReq.newBuilder()
-            .addAllEventTypes(eventTypes)
-            .setStartHeight(lastHeight + 1)
-            .setConsumer(StreamClientParams(uri, apiKey).consumer)
-            .build()
-        val responseObserver = EventStreamResponseObserver<EventProtos.EventBatch>() { batch ->
+
+        val responseObserver = EventStreamResponseObserver<EventBatch>() { batch ->
```
scirner22, line 91: That's cool that you transformed the json format into what we were already using here and it's even using the StreamObserver.

p8e-api/build.gradle
```diff
@@ -157,3 +181,41 @@ dependencyManagement {
         cacheChangingModulesFor 0, 'seconds'
     }
 }
+
+task downloadZipFile(type: Download) {
```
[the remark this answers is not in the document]
piercetrey-figure: hmm, good point. I'll do some tests/investigation to see what happens when there is no internet connection when building. Could maybe still leave it as an independent Gradle task and remove the generateProtoTasks dependency on it, and have it copy the protobufs into committed source instead of the build directory, so we can just run it whenever we need to update

p8e-api/src/main/kotlin/io/provenance/engine/stream/ScopeStream.kt
piercetrey-figure: yeah, seemed best to avoid as many changes here as possible at least for now... though I don't know if it is weird that the EventStreamResponseObserver still implements the grpc StreamObserver interface, given that this isn't interacting directly with grpc... I might remove that...

p8e-api/build.gradle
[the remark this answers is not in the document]
scirner22: That's a good idea.

rchaing-figure:
does brining in the protos makes p8e the owner?

piercetrey-figure:
> does brining in the protos makes p8e the owner?

@rchaing-figure no, they just aren't publishing these anywhere else yet, so we can do this as a periodic sync whenever they update

p8e-api/build.gradle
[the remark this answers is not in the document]
piercetrey-figure: I have this going now, as well as pulling in the base Cosmos SDK protos (the Gradle equivalent of this snippet Phil shared for getting the base cosmos protos `curl -L https://github.com/cosmos/cosmos-sdk/tarball/$(go list -mod=mod -m github.com/cosmos/cosmos-sdk | cut -d' ' -f2) | tar xvf - --strip-components 1 --include '*.proto'`), I think this has the base services that we will need for the full grpc switchover once we are able to develop against the new chain

iramiller:
> > does brining in the protos makes p8e the owner?
> 
> 
> 
> @rchaing-figure no, they just aren't publishing these anywhere else yet, so we can do this as a periodic sync whenever they update

The protos from provenance are in a zip file now with releases in the assets in GH.

piercetrey-figure:
@iramiller yes, I am downloading that zip of protos, as well as the base Cosmos protos via Gradle tasks

p8e-api/src/main/kotlin/io/provenance/engine/stream/EventStream.kt
```diff
@@ -0,0 +1,288 @@
...
+                    handleError(EventStreamStaleException("EventStream has not received a block in 30 seconds [lastBlockSeen: $lastBlockSeen vs. lastBlockMonitored: $lastBlockMonitored]"))
+                    break
+                }
+                lastBlockMonitored = lastBlockSeen
+            }
+            log.info("Exiting event monitor thread")
+        }
+
+        fun streamEvents() {
+            // todo: need to limit how many times this function is called??? Used to limit based on consumer id... probably need to use redis to do this... and ensure cleaned up when shutting down if do
```
piercetrey-figure, line 107: I think we probably want to do this (prevent an EventStream running simultaneously on multiple instances of p8e with the same consumer id so we don't double-process events when deploying, etc)...

the old event-stream could easily do this as it was just one node and kept track of the connected consumers, but we would either need to use Redis (which I know we have discussed getting rid of) or maybe have a heartbeat that touches some record in a db table signaling that it is still alive... thoughts on an approach here?

p8e-api/src/main/kotlin/io/provenance/engine/domain/Transaction.kt
```diff
@@ -0,0 +1,111 @@
+package io.provenance.engine.domain
+
+import com.fasterxml.jackson.annotation.JsonProperty
+import feign.Param
+import feign.RequestLine
+import io.provenance.engine.stream.domain.*
+import org.springframework.stereotype.Component
+
+data class BlockchainInfo(
```
piercetrey-figure, line 9: Note: I explored getting rid of these data classes I created and instead parsing the json-rpc call responses back to the new cosmos protobuf responses in order to more seamlessly prepare for when we are developing against the public chain. However, there was really no clean mapping of messages to do this conversion, there were slight differences all over the place (either properties nested at a different level or slightly differently named properties)... so it seemed like maybe more trouble than it is worth for a bit of short term gain

p8e-api/src/main/kotlin/io/provenance/engine/stream/EventStream.kt
rchaing-figure: I think adding a new component (Redis) fo a small piece would be overkill? I'd prefer the heartbeat idea with a flag somewhere. 

Does the LifecycleRegistery provide consumer information that we can utilize?

p8e-api/build.gradle
```diff
@@ -157,3 +181,64 @@ dependencyManagement {
         cacheChangingModulesFor 0, 'seconds'
     }
 }
+
+task downloadCosmosProtos(type: Download) {
```
rchaing-figure, line 185: maybe unclear to me, if we have gradle tasks to download protos, do we still need to commit the proto files below?

p8e-api/src/main/kotlin/io/provenance/engine/service/TransactionQueryService.kt
```diff
@@ -0,0 +1,27 @@
+package io.provenance.engine.service
+
+import io.provenance.engine.domain.*
+import org.springframework.stereotype.Component
+
+@Component
+class TransactionQueryService(val rpcClient: RPCClient) {
```
rchaing-figure, line 7: We can probably get a small unit test around this class? :badpokerface

piercetrey-figure: haha, I do have a test in the works for the EventStream itself, just haven't pushed it yet... I guess I could test that if the rpc sends us what it is supposed to we will pass it along 😉

p8e-api/src/main/kotlin/io/provenance/engine/service/TransactionQueryService.kt
```diff
@@ -0,0 +1,27 @@
+package io.provenance.engine.service
+
+import io.provenance.engine.domain.*
+import org.springframework.stereotype.Component
+
+@Component
```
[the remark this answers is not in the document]
piercetrey-figure: Maybe, does that set up the dependency injection the same way? If so I can change it, seems like we just @Component for everything, lol

p8e-api/build.gradle
piercetrey-figure: that's how I initially had it set up, but changed it because of @scirner22 's comment above https://github.com/provenance-io/p8e/pull/9#discussion_r595658957 ... with this way you don't have to be actively connected to the internet to build (so you can work while on an ✈️ I suppose)... but maybe I can find a way to cache the zip/tar to behave kind of like the other dependencies while offline, use the cache if present and/or download fails

p8e-api/src/main/kotlin/io/provenance/engine/stream/EventStream.kt
piercetrey-figure: I think the LifecycleRegistry is really just a state machine that allows to control when the WebSocket closes/opens, so I don't think it could facilitate communication between instances

p8e-api/src/main/kotlin/io/provenance/engine/domain/Transaction.kt
scirner22: Sounds good.

p8e-api/src/main/kotlin/io/provenance/engine/stream/EventStream.kt
scirner22: We can start keeping track of these as `TODO: concurrency-` or something that way we can find and solve for them at once.

p8e-api/src/main/kotlin/io/provenance/engine/stream/ScopeStream.kt
```diff
@@ -49,36 +49,9 @@ class ScopeStream(
     // The current event stream ID
     private val eventStreamId = java.util.UUID.fromString(eventStreamProperties.id)
 
-    // The event stream service uri
-    private val uri = URI(eventStreamProperties.uri)
```
scirner22, line 72: This reminded me we need to remove this from the properties files if it wasn't already 😉

p8e-api/src/main/kotlin/io/provenance/engine/stream/EventStream.kt
```diff
@@ -0,0 +1,288 @@
...
+            shutdownHook { shutdown() } // ensure we close socket gracefully when shutting down
+
+            timed("EventStream:streamHistory") {
+                streamHistory()
+            }
+
+            eventMonitor.start()
+        }
+
+        private fun streamHistory() {
```
scirner22, line 120: Does fetching the history here basically perform the same function as the tx reaper we have that would process skipped events?

scirner22:
Looks good! Mind performing a functional test where you try processing back from way back in time and you kill/restart the blockchain nodes a few times?

p8e-api/src/main/kotlin/io/provenance/engine/service/TransactionQueryService.kt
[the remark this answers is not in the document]
rchaing-figure: yea it should, only major difference would be components get precedence in initialization over service I believe in spring.

p8e-api/src/main/kotlin/io/provenance/engine/batch/TxErrorReaper.kt
```diff
@@ -46,47 +50,47 @@ class TxErrorReaper(
...
             }
         }
     }
 
-    private fun TxQuery.isErrored() = (code ?: 0) > 0
+    private fun GetTxResult.isNotFound() = txResult == null
+
+    private fun GetTxResult.isErrored() = txResult?.code != null && txResult?.code > 0
+
+    private fun GetTxResult.isSuccessful() = txResult?.code == 0
```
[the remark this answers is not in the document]
piercetrey-figure: I did an experiment just to confirm but null == 0 returns false, so it will still return false if there is no code, the isErrored function has to have the null check to make the comparison to 0 only happen if there is a code

p8e-api/src/main/kotlin/io/provenance/engine/batch/TxErrorReaper.kt
```diff
@@ -46,47 +50,47 @@ class TxErrorReaper(
...
 
-    private fun TxQuery.getError() = logs?.filter { it.log.isNotBlank() }?.takeIf { it.isNotEmpty() }?.joinToString("; ") { it.log } ?: raw_log
+    private fun GetTxResult.getError() = txResult?.log ?: "Unknown Error"
 
-    private fun TxQuery.events(): List<StdTxEvent>? = logs?.flatMap { it.events.filter { it.type.isScopeEventType() } }
+    private fun GetTxResult.scopeEvents(): List<Event>? = txResult?.events?.filter { it.type.isScopeEventType() }
```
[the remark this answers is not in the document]
piercetrey-figure: good call, pushed that up

p8e-api/src/main/kotlin/io/provenance/engine/stream/ScopeStream.kt
piercetrey-figure: oh yeah, additionally, I need to add in the other new urls to the non-local files

p8e-api/src/main/kotlin/io/provenance/engine/stream/EventStream.kt
piercetrey-figure: um, it is a bit different, this is streaming history from the last block we recorded to the current latest block, while the error reaper is a fallback to process blocks if we somehow skip over them and the event stream has continued on... it could be that now that we have the event stream in-house this could be changed around a bit, but for now they are doing separate things

p8e-api/src/main/kotlin/io/provenance/engine/service/TransactionQueryService.kt
piercetrey-figure: Took me a while to figure out a half reasonable way to mock events getting sent through the WebSocket, the documentation of Scarlet is no help, I ended up basically just mocking the RxKotlin Flowable responses from the Scarlet service and that seems to work pretty well... and the coverage is at least going up over all for this PR

image

p8e-api/src/test/kotlin/stream/EventStreamTest.kt
```diff
@@ -0,0 +1,180 @@
...
+        return EventStreamFactory.EventStream(eventTypes, startHeight, observer, lifecycle, eventStreamService, transactionQueryService)
+    }
+
+    @Test
+    fun `History is not streamed if startHeight is less than 0`() {
+        setLastBlockHeight(100)
+
+        buildEventStream(startHeight = -1).streamEvents()
+
+        verify(transactionQueryService, times(0)).blocksWithTransactions(any(), any())
```
rchaing-figure, line 52: Could we assert the return value? and maybe pass in real numbers?

piercetrey-figure: well in this case I am asserting that it wasn't ever called, so I think we would want the any argument and there is no return to assert...

rchaing-figure: sorry, i wasn't clear, but was speaking in general for this method call, i believe `blockWithTransactions` method returns a List of Longs, think we can assert as well that an empty list is returned or a long value? with non-any inputs

piercetrey-figure:
> Looks good! Mind performing a functional test where you try processing back from way back in time and you kill/restart the blockchain nodes a few times?

@scirner22 tried this, seemed to work well, the scheduled function in ScopeStream would exit due to the connection dropping, and then try every 30 seconds until up, and once up pick off streaming history where it left off.

p8e-api/src/main/kotlin/io/provenance/engine/stream/EventStream.kt
piercetrey-figure: added a concurrency note.

I looked into this a bit more and realized we were always using an empty string as our 'consumer id', so the event stream was actually considering us an anonymous consumer and not enforcing the single connection anyways, so leaving this as-is will be no different than it currently is

p8e-api/src/main/kotlin/io/provenance/engine/service/TransactionQueryService.kt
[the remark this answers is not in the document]
piercetrey-figure: changed

p8e-api/src/main/kotlin/io/provenance/engine/stream/EventStream.kt
scirner22: 👍

scirner22 (approved):

PR #11, opened 2021-03-23
Fix load proto memory class loader

scirner22:
Fix bug where load proto grpc could only load a class if it was already loaded into memory. This is because the class loader done inside of this function was using the proto jar instead of the contract jar. The proto jar is not an uberjar and thus won't have the runtime classpath.


piercetrey-figure (approved):
Nice, the fact that we can get rid of the whole child spec thing is really nice (assuming this works 🤣)

PR #12, opened 2021-03-23
working memorialize contract spec to new chain

piercetrey-figure:
- doc for protobuf serialization https://github.com/cosmos/cosmos-sdk/blob/master/docs/architecture/adr-027-deterministic-protobuf-serialization.md - I am just using the default serialization for the TxRaw serialization, but it seems to be conforming to this spec (I also tried their 'Article' example and the serialization for that matched)
- doc for proto tx signing https://github.com/cosmos/cosmos-sdk/blob/master/docs/architecture/adr-020-protobuf-transaction-encoding.md (we are using SIGN_MODE_DIRECT)
- also fixed an issue with EventStream where now txs_results on the block_results rpc endpoint apparently can be nullable now if there are no transactions vs. an empty array...


p8e-api/src/main/kotlin/io/provenance/engine/domain/Transaction.kt
```diff
@@ -61,7 +61,7 @@ class BlockResultsRequest(
 data class BlockResults(
     val height: Long,
     @JsonProperty("txs_results")
-    val txsResults: List<TxResult>
+    val txsResults: List<TxResult>?
```
piercetrey-figure, line 64: not sure if the spec changed so the endpoint that returns this can return null now instead of an empty array, but I was getting null locally and had to fix

p8e-api/src/main/kotlin/io/provenance/engine/service/ChaincodeInvokeService.kt
```diff
@@ -285,15 +288,20 @@ class ChaincodeInvokeService(
...
-                    }
-                }.also {
-                    scBatchTx(it).also {
-                        log.info("batch made it to mempool with txhash = ${it.txhash}")
+            val txBody = specList.map {
+                MsgAddP8eContractSpecRequest.newBuilder()
+                    .setContractspec(it.toProv())
+                    .addAllSigners(listOf(accountProvider.bech32Address()))
+                    .build()
+            }.toTxBody()
```
piercetrey-figure, line 296: added this helper `toTxBody` to easily take a list of messages or an individual message and get the TxBody object with properly packed Any protos...

@scirner22 should be useful once you have a MsgP8eMemorializeContractRequest built

p8e-api/src/main/kotlin/io/provenance/engine/service/ProvenanceGrpcService.kt
```diff
@@ -0,0 +1,147 @@
...
+    private val bech32Address = accountProvider.bech32Address()
+    private val keyPair = accountProvider.getKeyPair()
+    private val signer = PbSigner.signerFor(keyPair)
+
+    fun accountInfo(): Auth.BaseAccount = accountService.account(QueryOuterClass.QueryAccountRequest.newBuilder()
+            .setAddress(bech32Address)
+            .build()
+        ).run { account.unpack(Auth.BaseAccount::class.java) }
+
+    fun signTx(body: TxBody, accountInfo: Auth.BaseAccount, sequenceNumberOffset: Int = 0, gasEstimate: GasEstimate = GasEstimate(0)): Tx {
```
piercetrey-figure, line 63: Note: I haven't tested the sequence number and offset, once that is all converted though for the memorialize contract, I would expect it to work fine if you stuff a bunch of transactions inside a block following the logic we currently use with SimpleClient

iramiller: In case this comes up... one signed tx (sequence number) with many messages... the terminology can get confusing on the blockchain side because we think about submitting many requests as many transactions... this point becomes important though because all those messages succeed or fail as a group. You can submit more than one tx per block (potentially again with more than one msg each) if you bump the sequence number each time you sign.

p8e-api/src/main/kotlin/io/provenance/engine/service/ChaincodeInvokeService.kt
```diff
@@ -40,242 +55,268 @@ class ChaincodeInvokeService(
...
+                        val message = messages.pop()
+
+                        if (messages.isEmpty()) {
+                            log.debug("removing scope $key from backlog as its list is now empty")
+
+                            priorityScopeBacklog.remove(key)
+                        }
 
-            val batch = mutableMapOf<UUID, BlockchainTransaction>()
+                        blockScopeIds.add(message.request.scopeId)
```
piercetrey-figure, line 98: probably should have a matching debug message to the one in the queue.poll block below (or move the message and blockScopeIds/batch.add to a separate function)

p8e-api/src/main/kotlin/io/provenance/engine/service/ChaincodeInvokeService.kt
```diff
@@ -11,21 +14,33 @@ import io.p8e.util.toUuidProv
...
+import java.util.concurrent.Future
+import java.util.concurrent.LinkedBlockingQueue
+import java.util.concurrent.TimeUnit
 import kotlin.concurrent.thread
+import kotlin.random.Random
+
+const val SIGNATURE_VERIFICATION_FAILED = 4
+
+data class ContractRequestWrapper(
+    val uuid: UUID,
```
piercetrey-figure, line 39: maybe should rename this executionUuid, since that appears to be what we are putting here... seems like we have too many uuids to not be specific =)

p8e-shared/src/main/kotlin/io/provenance/p8e/shared/domain/Envelope.kt
```diff
@@ -72,6 +72,10 @@ open class EnvelopeEntityClass : UUIDEntityClass<EnvelopeRecord>(
         EnvelopeTable.groupUuid eq groupUuid and (EnvelopeTable.executionUuid eq executionUuid)
     }.toList()
 
+    fun findByExecutionUuid(executionUuid: UUID) = find {
```
rchaing-figure, line 75: Shouldn't we find with public key? for PII purposes.

piercetrey-figure: No, because we need to get all parties envelopes for this execution that exist on a single p8e instance, so they can all proceed with being completed (after we index the scope)

p8e-api/src/main/kotlin/io/provenance/engine/service/ProvenanceGrpcService.kt
```diff
@@ -0,0 +1,192 @@
...
+            NettyChannelBuilder.forAddress(uri.host, uri.port)
+                .also {
+                    if (uri.scheme == "grpcs") {
+                        it.useTransportSecurity()
+                    } else {
+                        it.usePlaintext()
+                    }
+                }
+                // TODO try default size of 3MB for now
+                .maxInboundMessageSize(40 * 1024 * 1024) // ~ 20 MB
```
scirner22, line 58: we'll have to see how 40mb does in testnet 😄

scirner22 (approved):

PR #13, opened 2021-03-29
Fix/multiparty pre execution

scirner22:
Fixes two bugs that were noticed recently:
- Setting `this.executionEnvelope` can fail after the atomic boolean is already set to true. When it is retried it will be short circuited and an empty envelope is submitted to the backend. It ends up failing for being empty.
- Cross scope facts were being fetched as protos and then saved back to object store from protos. This is needed to permission cross scope facts for recitals that are brand new to the scope. There's an edge case where the bytes into a proto is different than the bytes out of a proto. This leads to new recital members not getting access to the exact hash they should even though the protos remain equivalent in this case.

### DIFF
--- a/p8e-sdk/src/main/kotlin/io/p8e/proxy/PermissionUpdater.kt
+++ b/p8e-sdk/src/main/kotlin/io/p8e/proxy/PermissionUpdater.kt
@@ -4,6 +4,7 @@ import com.google.protobuf.Message
 import io.p8e.ContractManager
 import io.p8e.proto.Contracts.Contract
 import io.p8e.util.ThreadPoolFactory
+import java.io.ByteArrayInputStream
 import java.security.PublicKey
 import java.util.UUID
 import java.util.concurrent.CompletableFuture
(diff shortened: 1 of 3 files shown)


rchaing-figure (approved):

p8e-sdk/src/main/kotlin/io/p8e/proxy/PermissionUpdater.kt
```diff
@@ -17,12 +18,16 @@ class PermissionUpdater(
...
+        // the recitals
         contract.inputsList.threadedMap(executor) { fact ->
-            val message = contractManager.loadProto(
-                fact.dataLocation.ref.hash,
-                fact.dataLocation.classname
-            )
-            contractManager.saveProto(message, audience = audience)
+            with (contractManager.client) {
+                val obj = this.loadObject(fact.dataLocation.ref.hash)
+                val inputStream = ByteArrayInputStream(obj)
```
rchaing-figure, line 27: we dont care about classname?

scirner22: All the classname does it cast the loaded proto. In this case we are now loading the bytes and saving the bytes directly so we never need to cast to the classname.

PR #18, opened 2021-04-04
Rework contract manager watcher algorithm

scirner22:
# Background

There was a reported issue where a watcher was not reconnecting. To an end user when this case happens it appears that all contracts are hanging.

# Overview

Moves watcher disconnect handler to a scheme where we are constantly checking the desired and actual state. This allows for envelope observer to only last for a single stream rpc and a new one is created for the next iteration. This is an improvement over the previous logic where there was a retry handler built into the stream observer itself. This allows the algorithm that manages a constant connection to be simpler.

## Notes

- The disconnect handler is now deprecated. As part of this change it was determined that when you start a watcher the goal for it is to remain "watching" until `unwatch(...)` or `close(...)` is called. Because of this the retry is built in for all cases by default.


p8e-sdk/src/main/kotlin/io/p8e/ContractManager.kt
```diff
@@ -144,8 +147,10 @@ class ContractManager(
         val errorHandler: ContractErrorHandler<T>
     )
 
-    private val heartbeatConnections = ConcurrentHashMap<HeartbeatConnectionKey, HeartbeatQueuer>()
```
scirner22, line 147: This could have been the source of some of the problems. The heartbeatConnections held a reference to the queuer. This means that it's possible the queuer could have been out of sync from the queuers array and the heartbeat was pinging a queuer that wasn't present in the queuers map. This would mean it would be connected but couldn't receive or process events.

p8e-sdk/src/main/kotlin/io/p8e/async/Heartbeat.kt
```diff
@@ -0,0 +1,111 @@
...
+                                io.grpc.Status.UNAVAILABLE.code -> true
+                                else -> false
+                            }
+                        } else {
+                            false
+                        }
+                    }
+                }
+            }?.run {
+                log.info("Sleeping for 10 seconds since last batch contains a network disconnect")
```
piercetrey-figure, line 84: why do we need to sleep in this case?

piercetrey-figure (approved):

p8e-sdk/src/main/kotlin/io/p8e/async/Heartbeat.kt
scirner22: Only because when there's a failure due to a network problem, odds are it will take more than a second for it to be fixed. Without this the logs just get spammed more and it retries quite often.

rchaing-figure (approved):

PR #20, opened 2021-04-08
Don't push to latest tag on prereleases.

scirner22:
## Description

As it currently stands we are pushing to docker `:latest` tag on prereleases.


rchaing-figure (approved):

scirner22:
Confirmation of no latest tag on the most recent pre-release.

PR #21, opened 2021-04-08
Allow key and value in tx result from provenance grpc to be nullable.

scirner22:
## Description

```
Caused by: feign.FeignException: Instantiation of [simple type, class io.provenance.engine.stream.domain.Attribute] value failed for JSON property value due to missing (therefore NULL) value for creator parameter value which is a non-nullable type

 at [Source: (BufferedReader); line: 38, column: 15] (through reference chain: io.provenance.engine.stream.domain.RPCResponse["result"]->io.provenance.engine.domain.BlockResults["txs_results"]->java.util.ArrayList[0]->io.provenance.engine.domain.TxResult["events"]->java.util.ArrayList[1]->io.provenance.engine.stream.domain.Event["attributes"]->java.util.ArrayList[2]->io.provenance.engine.stream.domain.Attribute["value"]) reading GET http://35.190.171.250:26657/

	at feign.FeignException.errorReading(FeignException.java:151)

	at feign.AsyncResponseHandler.handleResponse(AsyncResponseHandler.java:102)

	at feign.SynchronousMethodHandler.executeAndDecode(SynchronousMethodHandler.java:138)

	at feign.SynchronousMethodHandler.invoke(SynchronousMethodHandler.java:89)

	at feign.ReflectiveFeign$FeignInvocationHandler.invoke(ReflectiveFeign.java:100)

	at com.sun.proxy.$Proxy116.blockResults(Unknown Source)
```

### DIFF
--- a/p8e-api/src/main/kotlin/io/provenance/engine/domain/Transaction.kt
+++ b/p8e-api/src/main/kotlin/io/provenance/engine/domain/Transaction.kt
@@ -67,7 +67,7 @@ data class BlockResults(
 data class TxResult(
     val code: Int?,
     val data: String?,
-    val log: String,
+    val log: String?,
     val info: String,
     val gasWanted: Long,
     val gasUsed: Long,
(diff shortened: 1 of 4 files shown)


p8e-api/src/main/kotlin/io/provenance/engine/domain/Transaction.kt
```diff
@@ -67,7 +67,7 @@ data class BlockResults(
 data class TxResult(
     val code: Int?,
     val data: String?,
-    val log: String,
+    val log: String?,
```
scirner22, line 70: I saw the simple client allows this to be nullable so I added it here as well.

rchaing-figure (approved):

piercetrey-figure (approved):

PR #23, opened 2021-04-14
Create an abstract signer class

rchaing-figure:
Add a Signer base class that implicitly references the key management class for signing.

- [x] Test that Pen signer still works as the default. 

~~ **PR'ing against a feature branch, so that we don't have a partially working feature.** ~~

### DIFF
--- a/p8e-api/src/main/kotlin/io/provenance/engine/config/AppConfig.kt
+++ b/p8e-api/src/main/kotlin/io/provenance/engine/config/AppConfig.kt
@@ -8,9 +8,10 @@ import com.tinder.scarlet.messageadapter.moshi.MoshiMessageAdapter
 import com.tinder.scarlet.streamadapter.rxjava2.RxJava2StreamAdapterFactory
 import com.tinder.scarlet.websocket.okhttp.newWebSocketFactory
 import feign.Feign
-import feign.Logger
 import feign.jackson.JacksonDecoder
 import feign.jackson.JacksonEncoder
+import io.p8e.crypto.Pen
+import io.p8e.crypto.SmartKeySigner
 import io.p8e.util.configureProvenance
 import io.provenance.engine.crypto.Account
 import io.provenance.engine.crypto.PbSigner
@@ -77,6 +78,8 @@ import java.time.Duration
     ServiceProperties::class,
     ProvenanceKeystoneProperties::class,
     MetricsProperties::class,
+    SmartKeyProperties::class,
+    SignerProperties::class
 ])
 class AppConfig : WebMvcConfigurer {
 
@@ -260,5 +263,16 @@ class AppConfig : WebMvcConfigurer {
                     first to second
                 }.toMap()
             MetricsService(collectors, labels)
+        }
+
+    /**
+     * Add support for new key management.
+     */
+    @Bean
+    fun signer(smartKeyProperties: SmartKeyProperties, signerProperties: SignerProperties) =
+       when(signerProperties.type.toLowerCase()) {
+            "default" -> Pen()
+            "smartkey" -> SmartKeySigner(smartKeyProperties.apiKey)
+            else -> throw UnsupportedOperationException("Unsupported key management type ${signerProperties.type}")
         }
 }
--- a/p8e-api/src/main/kotlin/io/provenance/engine/config/AppConfig.kt
+++ b/p8e-api/src/main/kotlin/io/provenance/engine/config/AppConfig.kt
@@ -10,7 +10,8 @@ import com.tinder.scarlet.websocket.okhttp.newWebSocketFactory
 import feign.Feign
 import feign.jackson.JacksonDecoder
 import feign.jackson.JacksonEncoder
-import io.p8e.crypto.Pen
+import io.p8e.crypto.SignerFactory
+import io.p8e.crypto.SignerImpl
 import io.p8e.crypto.SmartKeySigner
 import io.p8e.util.configureProvenance
 import io.provenance.engine.crypto.Account
@@ -78,8 +79,7 @@ import java.time.Duration
     ServiceProperties::class,
     ProvenanceKeystoneProperties::class,
     MetricsProperties::class,
-    SmartKeyProperties::class,
-    SignerProperties::class
+    SmartKeyProperties::class
 ])
 class AppConfig : WebMvcConfigurer {
 
@@ -269,10 +269,7 @@ class AppConfig : WebMvcConfigurer {
      * Add support for new key management.
      */
     @Bean
-    fun signer(smartKeyProperties: SmartKeyProperties, signerProperties: SignerProperties) =
-       when(signerProperties.type.toLowerCase()) {
-            "default" -> Pen()
-            "smartkey" -> SmartKeySigner(smartKeyProperties.apiKey)
-            else -> throw UnsupportedOperationException("Unsupported key management type ${signerProperties.type}")
-        }
+    fun signer(smartKeyProperties: SmartKeyProperties): SignerFactory {
+        return SignerFactory(smartKeyProperties.apiKey)
+    }
 }
--- a/p8e-api/src/main/kotlin/io/provenance/engine/config/Properties.kt
+++ b/p8e-api/src/main/kotlin/io/provenance/engine/config/Properties.kt
@@ -140,8 +140,3 @@ class MetricsProperties {
 class SmartKeyProperties {
     @NotNull lateinit var apiKey: String
 }
-
-@ConfigurationProperties(prefix = "signer")
-class SignerProperties {
-    @NotNull var type: String = "default" // Pen class default signer
-}
--- a/p8e-api/src/main/kotlin/io/provenance/engine/config/Properties.kt
+++ b/p8e-api/src/main/kotlin/io/provenance/engine/config/Properties.kt
@@ -135,3 +135,13 @@ class MetricsProperties {
     var tags: String = ""
     var prefix: String = "p8e_api"
 }
+
+@ConfigurationProperties(prefix = "smartkey")
+class SmartKeyProperties {
+    @NotNull lateinit var apiKey: String
+}
+
+@ConfigurationProperties(prefix = "signer")
+class SignerProperties {
+    @NotNull var type: String = "default" // Pen class default signer
+}
--- a/p8e-proto-internal/src/main/proto/p8e/common.proto
+++ b/p8e-proto-internal/src/main/proto/p8e/common.proto
@@ -17,6 +17,8 @@ message Signature {
 
     // Identity of signer
     SigningAndEncryptionPublicKeys signer = 4;
+
+    UUID key_id = 5;
 }
 
 message BooleanResult {
--- a/p8e-util/src/main/kotlin/io/p8e/crypto/Pen.kt
+++ b/p8e-util/src/main/kotlin/io/p8e/crypto/Pen.kt
@@ -50,12 +50,12 @@ class Pen: SignerImpl() {
             SIGN_ALGO,
             PROVIDER
         )
-        s.initSign(keys?.private)
+        s.initSign(keys!!.private)
         s.update(data)
 
         return ProtoUtil
             .signatureBuilderOf(String(s.sign().base64Encode()))
-            .setSigner(lens?.signer())
+            .setSigner(lens!!.signer())
             .build()
             .takeIf { lens!!.verify(data, it) }
             .orThrow { IllegalStateException("can't verify signature - public cert may not match private key.") }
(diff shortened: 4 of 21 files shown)


p8e-api/src/main/kotlin/io/provenance/engine/config/Properties.kt
```diff
@@ -135,3 +135,13 @@ class MetricsProperties {
     var tags: String = ""
     var prefix: String = "p8e_api"
 }
+
+@ConfigurationProperties(prefix = "smartkey")
+class SmartKeyProperties {
+    @NotNull lateinit var apiKey: String
```
rchaing-figure, line 141: Considered adding a boolean flag for enable/disable, but didn't make sense if we set a flag for each of the signer then developer would expect that it would write to both places, which we want to avoid.

p8e-api/src/main/kotlin/io/provenance/engine/config/AppConfig.kt
```diff
@@ -261,4 +264,15 @@ class AppConfig : WebMvcConfigurer {
...
         }
+
+    /**
+     * Add support for new key management.
+     */
+    @Bean
+    fun signer(smartKeyProperties: SmartKeyProperties, signerProperties: SignerProperties) =
+       when(signerProperties.type.toLowerCase()) {
+            "default" -> Pen()
+            "smartkey" -> SmartKeySigner(smartKeyProperties.apiKey)
```
rchaing-figure, line 275: Will improve the signer selection here with enum type, the SmartKeySigner() does not work with p8e just yet.

p8e-api/src/main/kotlin/io/provenance/engine/config/Properties.kt
```diff
@@ -135,3 +135,13 @@ class MetricsProperties {
...
+}
+
+@ConfigurationProperties(prefix = "signer")
+class SignerProperties {
+    @NotNull var type: String = "default" // Pen class default signer
```
rchaing-figure, line 146: Need to set on a enum type.

p8e-proto-internal/src/main/proto/p8e/common.proto
```diff
@@ -17,6 +17,8 @@ message Signature {
 
     // Identity of signer
     SigningAndEncryptionPublicKeys signer = 4;
+
+    UUID key_id = 5;
```
rchaing-figure, line 21: SmartKey has a key_id as part of the signature that we need we need to verify the signature.

p8e-util/src/main/kotlin/io/p8e/crypto/SignerImpl.kt
```diff
@@ -0,0 +1,22 @@
+package io.p8e.crypto
+
+import com.google.protobuf.Message
+import io.p8e.proto.Common.Signature
+import java.security.KeyPair
+
+open class SignerImpl {
```
rchaing-figure, line 7: Considering implementing the Pen class in here and label the SignerImpl class as the default, but wanted to separate both the signers for now so it easier to visualize and test.

p8e-util/src/main/kotlin/io/p8e/crypto/SmartKeySigner.kt
```diff
@@ -0,0 +1,71 @@
...
+import com.fortanix.sdkms.v1.model.DigestAlgorithm
+import com.fortanix.sdkms.v1.model.SignRequest
+import com.fortanix.sdkms.v1.model.VerifyRequest
+import com.google.protobuf.Message
+import io.p8e.proto.Common.Signature
+import io.p8e.util.orThrow
+import io.p8e.util.toProtoUuidProv
+import java.lang.IllegalStateException
+
+class SmartKeySigner(
```
rchaing-figure, line 17: This work as a standalone already, need to integrate with p8e.

p8e-api/src/main/kotlin/io/provenance/engine/config/AppConfig.kt
```diff
@@ -261,4 +264,15 @@ class AppConfig : WebMvcConfigurer {
                 }.toMap()
             MetricsService(collectors, labels)
         }
+
+    /**
+     * Add support for new key management.
+     */
+    @Bean
+    fun signer(smartKeyProperties: SmartKeyProperties, signerProperties: SignerProperties) =
```
scirner22, line 272: Since this signer is in a bean and in order to sign with it you use `setKeyId(..)` first doesn't this produce a race condition where two concurrent requests from different affiliates won't be signed correctly? Using this pattern you have I think you want this bean to return a SignerFactory.

p8e-proto-internal/src/main/proto/p8e/common.proto
scirner22: Can we get away with not having this? When you upload or create a new key in smart key it lets you download the public key. We can store that public key the same way we do in the affiliates table now. It would just also have the smart_key_uuid on the row as well.

scirner22 (changes requested):
Nice idea with the feature branch. That's a good idea we can make more use of.

p8e-api/src/main/kotlin/io/provenance/engine/config/AppConfig.kt
rchaing-figure: > I think you want this bean to return a SignerFactory.
Yea, think this what it does, in this case it's returning the "SignerImpl" class, both signer inherits from it. 

> first doesn't this produce a race condition where two concurrent requests from different affiliates won't be signed correctly?
I'd have to test this more to be absolutely sure, but its highly possible now that I think about it, personally I don't like the idea of having to `setKeyId` at the time of handling, I'd probably want the affiliateService to handle setting this to prevent any possible race condition.

p8e-proto-internal/src/main/proto/p8e/common.proto
rchaing-figure: Yea, I could make the change along with this ticket ... [https://app.clubhouse.io/figure/story/88583/update-affiliate-table-to-include-securityobject-table-uuid](Update affiliate table to include SecurityObject Table UUID), I've made a note to remove it.

p8e-util/src/main/kotlin/io/p8e/crypto/SmartKeySigner.kt
```diff
@@ -0,0 +1,76 @@
...
+
+        val signatureResponse = SignAndVerifyApi().sign(sObjUUID, signatureRequest)
+
+        return Signature.newBuilder()
+            .setAlgo(DigestAlgorithm.SHA512.value)
+            .setProvider("SmartKey")
+            .setSignature(signatureResponse.signature.toString())
+            .setKeyId(signatureResponse.kid.toProtoUuidProv())
+            .build()
+            .takeIf { verify(data, it) }
```
piercetrey-figure, line 61: do we really need to verify every time we sign?

p8e-api/src/main/kotlin/io/provenance/engine/config/AppConfig.kt
piercetrey-figure: I second the idea of the bean returning a factory that can return a new instance of a SignerImpl based on what is passed in (i.e. a uuid or a keyPair) so multiple signers can operate independently

p8e-util/src/main/kotlin/io/p8e/crypto/Pen.kt
```diff
@@ -20,40 +20,48 @@ class Pen(privateKey: PrivateKey, publicKey: PublicKey) {
...
+    var keys: KeyPair? = null
+    var lens: Lens? = null
+
+    override fun setKeyId(keyPair: KeyPair) {
+        keys = keyPair
+        lens = Lens(keys!!.public)
+    }
+
+    override fun setKeyId(uuid: String) {
+        /*no-op*/
```
piercetrey-figure, line 36: I wonder if it would be worth just populating the uuid on the affiliate table even when using the default driver and just fetch the keypair from the db in here so we could always use a uuid externally?

rchaing-figure: Yep, I have a ticket to remove passing keyPairs around and instead each keypair is referenced by a UUID and we can pull it that way.

p8e-util/src/main/kotlin/io/p8e/crypto/SmartKeySigner.kt
rchaing-figure: Not sure, I just copied this from the Pen class over, so that each of the signer looks identical, but I don't see any harm in once-overing the signature?

piercetrey-figure (approved):
Looks good once that race condition potential is thought through

p8e-util/src/main/kotlin/io/p8e/crypto/SmartKeySigner.kt
piercetrey-figure: I guess it is just one more request against our account. Either way we have to trust that what they send back is accurate, whether it be a signature or a validation result 😆

p8e-util/src/main/kotlin/io/p8e/crypto/Pen.kt
iramiller: It is maybe worth considering that instead of a uuid the address of the key should be used instead because unlike a uuid it is directly computed from the key... also everything blockchain based will use an address... ¯\_(ツ)_/¯

iramiller:
It is maybe worth considering that instead of a uuid the address of the key should be used instead because unlike a uuid it is directly computed from the key... also everything blockchain based will use an address... ¯\_(ツ)_/¯

p8e-util/src/main/kotlin/io/p8e/crypto/Pen.kt
piercetrey-figure: Aren't you on vacation @iramiller? 😄

rchaing-figure: Agreed, something worth considering now, this might require some conversion effort of keys within p8e cause not sure it makes sense to store the actual key + the address of that key. 

_**I shouldn't answer reply to someone that is vacationing wrong**_

p8e-api/src/main/kotlin/io/provenance/engine/config/AppConfig.kt
rchaing-figure: Think I get what you guys are saying, I just need to make what I have smarter.

p8e-util/src/main/kotlin/io/p8e/crypto/SignerFactory.kt
```diff
@@ -0,0 +1,20 @@
...
+
+import java.lang.IllegalStateException
+import java.security.KeyPair
+
+class SignerFactory(
+    smartKeyApiKey: String
+){
+    private val appApiKey = smartKeyApiKey
+
+    fun getSigner(uuid: String? = null, keyPair: KeyPair? = null): SignerImpl {
```
scirner22, line 11: this could be more kotlin idiomatic with a sealed class. For instance the idea would look like this.

```kotlin
sealed class Signer
data class KeyPairSigner(val keyPair: KeyPair) : Signer
data class SmartKeySigner(val uuid: String) : Signer

fun getSigner(signer: Signer): SignerImpl {
    return when (signer) {
        is KeyPairSigner -> Pen(signer.keyPair)
        is SmartKeySigner -> SmartKeySigner(appApiKey, signer.uuid)
    }
}
```

p8e-util/src/main/kotlin/io/p8e/crypto/SmartKeySigner.kt
scirner22: Since we have the public key from smart key, we should be able to sign locally and not have to use the api for it.

p8e-util/src/main/kotlin/io/p8e/crypto/SignerFactory.kt
rchaing-figure: i had something similar last night, in my mind i thought was this was like explicitly telling which signer to use, ill put up what i had back.

p8e-util/src/main/kotlin/io/p8e/crypto/SmartKeySigner.kt
rchaing-figure: Yea, we can do that as well, I'll handle it in the next branch.

p8e-util/src/main/kotlin/io/p8e/crypto/Pen.kt
piercetrey-figure: do we generate this uuid, or does that just come straight from SmartKey when we generate a key? I.e. we might not even control this?

p8e-util/src/main/kotlin/io/p8e/crypto/SmartKeySigner.kt
piercetrey-figure: ⭐  for @scirner22

p8e-util/src/main/kotlin/io/p8e/crypto/Pen.kt
rchaing-figure: they provide us a UUID for a SecurityObject (a locker for keys) is my real world analogy.

rchaing-figure:
merging into feature branch, we can continue to review over there if necessary.

PR #24, opened 2021-04-15
Signer Management Feature

rchaing-figure:
Signer Management Feature - Adding the flexibility for p8e to securely manage signing keys through a 3rd party or library.


piercetrey-figure:
@rchaing-figure I think we have to keep the protos under the objectstore directory, having them not nested caused issues for Peter (fixed previously here https://github.com/provenance-io/p8e/pull/56)

p8e-api/src/main/kotlin/io/provenance/engine/service/AuthenticationService.kt
```diff
@@ -50,8 +51,9 @@ class AuthenticationService(
             return null
         }
 
+        // Verify signature against the auth public key we have on record.
         val verified = Signature.getInstance(Constant.JWT_ALGORITHM).apply {
-            initVerify(publicKey)
+            initVerify(affiliate.authPublicKey.toJavaPublicKey())
```
piercetrey-figure, line 56: Just confirming that we are just going to copy existing affiliate's signing public key over to the auth_public_key column, right?

rchaing-figure: :yes

p8e-encryption/src/main/kotlin/io/provenance/p8e/encryption/ecies/ProvenanceECIESCipher.kt
```diff
@@ -0,0 +1,109 @@
...
+     * @throws ProvenanceECIESDecryptException In case decryption fails due to invalid key or invalid MAC value.
+     * @throws IllegalArgumentException In case decryption fails due to invalid life-cycle phase,
+     */
+    @Throws(ProvenanceECIESDecryptException::class)
+    fun decrypt(payload: ProvenanceECIESCryptogram, keyRef: KeyRef, additionalAuthenticatedData: String?): ByteArray {
+        try {
+
+            val ephemeralDerivedSecretKey = if(keyRef.type == SMARTKEY) {
+                // Create a transient security object out of the ephemeral public key from the payload.
+                val transientEphemeralSObj = payload.ephemeralPublicKey.toTransientSecurityObject()
```
piercetrey-figure, line 76: So this creates the transient security object, and it lasts just through the request in toAgreeKey on the next line of code?

piercetrey-figure: I'm assuming there is no way to make that in one request instead of two somehow?

rchaing-figure: yea, TTL is during this session, don't think I can make one request out of two, need the transient SObj to exist before we can transform it into an AgreeKey

p8e-api/src/main/kotlin/io/provenance/engine/service/MailboxService.kt
```diff
@@ -31,35 +31,37 @@ class MailboxService(
...
-            log.error("Invoker publicKey: ${invokerPublicKey.toHex()} does not match application public key: ${encryptionKeyPair.public.toHex()}")
+        if (invokerPublicKey != encryptionKeyRef.publicKey && invokerPublicKey != signer.getPublicKey()) {
+            log.error("Invoker publicKey: ${invokerPublicKey.toHex()} does not match application public key: ${encryptionKeyRef.publicKey.toHex()}")
             return
         }
 
         val scopeOwners = env.scope.partiesList
             .filter { it.hasSigner() }
             .map { it.signer.encryptionPublicKey.toPublicKey() }
+            .map { affiliateService.getEncryptionKeyRef(it) }
```
piercetrey-figure, line 48: couldn't some of these parties/recitals be affiliates on other p8e instances (and therefore the affiliateService wouldn't be able to fetch an affiliate from our db)? I'm wondering if the getEncryptionKeyRef should be able to fall back to a KeyRef containing only the public key in that case?

p8e-encryption/src/main/kotlin/io/provenance/p8e/encryption/ecies/ProvenanceECIESCipher.kt
piercetrey-figure: I would assume if the auth token refreshed between these two calls, that would still be the same 'session'? 🤞

piercetrey-figure (approved):
Looking good, just the one question on external affiliates

p8e-api/src/main/kotlin/io/provenance/engine/service/MailboxService.kt
rchaing-figure: hmm ... this is a good point, ill need to circle around and fix this.

p8e-encryption/src/main/kotlin/io/provenance/p8e/encryption/ecies/ProvenanceECIESCipher.kt
rchaing-figure: yea, the refresh extends the session

build.gradle
```diff
@@ -49,6 +49,8 @@ allprojects {
         implementation 'org.slf4j:log4j-over-slf4j:1.7.30'
         implementation "javax.inject:javax.inject:1"
 
+        implementation "com.fortanix:sdkms-client:3.23.1408"
```
scirner22, line 52: if this was moved only to sub projects that needed it, which projects would it be in? If we can keep it out of p8e-sdk and anything p8e-sdk needs that would be great to keep the client jar size down.

rchaing-figure: - p8e-util (where signing happens)
- p8e-encryption (decryption)
- p8e-api 
- p8e-api-webservice

p8e-sdk depends on p8e-util and p8e-encryption

p8e-api-webservice/src/main/kotlin/io/provenance/p8e/webservice/config/AppConfig.kt
```diff
@@ -78,4 +89,42 @@ class AppConfig : WebMvcConfigurer {
...
+                auth.apiKey = authResponse.accessToken
+                auth.apiKeyPrefix = "Bearer"
+            }
+        } else {
+            null // SmartKey is not initialized.
+        }
+    }
+
+    @Bean
+    fun authenticationApi(smartKeyApiClient: ApiClient): AuthenticationApi = AuthenticationApi(smartKeyApiClient)
```
scirner22, line 116: I guess these beans below don't complain when smartKeyApiClient is null?

p8e-api/src/main/kotlin/io/provenance/engine/batch/MailboxReaper.kt
```diff
@@ -240,7 +241,10 @@ class MailboxReaper(
                 }
             }.mapNotNull { (keyPair, result) ->
                 val (uuid, dimeInputStream) = result
-                dimeInputStream.getDecryptedPayload(keyPair).use {
+                val signer = transaction{ affiliateService.getSigner(keyPair.public) }
+                val encryptionKeyRef = transaction{ affiliateService.getEncryptionKeyRef(keyPair.public) }
+
+                dimeInputStream.getDecryptedPayload(encryptionKeyRef, signer).use {
```
scirner22, line 247: Does this work in all cases? I think the one parameter function `getDecryptedPayload(keyPair)` would use the first signer to verify. Passing an explicit signer would mean it fails verification if that signer didn't sign? Wouldn't that be the case for mailbox items because it's the sender that signed?

p8e-api/src/main/kotlin/io/provenance/engine/config/AppConfig.kt
```diff
@@ -210,4 +221,42 @@ class AppConfig : WebMvcConfigurer {
...
+                auth.apiKey = authResponse.accessToken
+                auth.apiKeyPrefix = "Bearer"
+            }
+        } else {
+            null // SmartKey is not initialized.
+        }
+    }
+
+    @Bean
+    fun authenticationApi(smartKeyApiClient: ApiClient): AuthenticationApi = AuthenticationApi(smartKeyApiClient)
```
scirner22, line 248: likewise with these calls like in the p8e-webservice config?

p8e-api/src/main/kotlin/io/provenance/engine/grpc/observers/EnvelopeEventObserver.kt
```diff
@@ -190,16 +190,17 @@ class EnvelopeEventObserver(
 
                 affiliateConnection.connectionStatus = CONNECTED
                 affiliateConnection.lastHeartbeat = OffsetDateTime.now()
-            }
```
scirner22, line 193: was this transaction extended just to get `affiliateRecord.publicKey.value.toPublicKeyProto()` below?

p8e-api/src/main/kotlin/io/provenance/engine/grpc/v1/ObjectGrpc.kt
```diff
@@ -43,24 +44,31 @@ class ObjectGrpc(
...
+        val encryptionKeyRef = transaction { affiliateService.getEncryptionKeyRef(publicKey()) }
 
-        // Update the dime's audience list to use encryption public keys.
+        // Update the dime's audience list to use the encryption key ref, so we has accessibility
+        // to the audience's UUID as well, if SmartKey is used.
         val audience = request.toAudience().plus(affiliateShares.value).map {
             transaction {
                 try {
-                    affiliateService.getEncryptionKeyPair(it).public
+                    affiliateService.getEncryptionKeyRef(it)
```
scirner22, line 56: Can't this now be getEncryptionPublicKeys() since we changed the encryption side to not need the private keys at all?

p8e-api/src/main/kotlin/io/provenance/engine/service/MailboxService.kt
scirner22: Is the encryption key ref needed at all? Shouldn't the public key be enough.

p8e-shared/src/main/kotlin/io/provenance/p8e/shared/service/AffiliateService.kt
```diff
@@ -53,17 +58,58 @@ class AffiliateService(
...
+     *  [signer] The specific signer used to sign contracts
+     */
+    fun getSigner(publicKey: PublicKey): SignerImpl {
+        val affiliateRecord = get(publicKey)
+        return if(affiliateRecord?.keyType == SMARTKEY) {
+            signerFactory.getSigner(SignerFactoryParam.SmartKeyParam(affiliateRecord.signingKeyUuid.toString()))
+        } else {
+            signerFactory.getSigner(
+                SignerFactoryParam.PenParam(
+                    KeyPair(affiliateRecord?.publicKey?.value?.toJavaPublicKey(), affiliateRecord?.privateKey?.toJavaPrivateKey())
```
scirner22, line 78: right here would probably be the best place to use `!!` since we know they should exist in the database key type or else we could throw an error saying there's a `DATABASE` affiliate with missing fields. That should help so we don't need to `!!` across other places where the signer is used.

p8e-shared/src/main/kotlin/io/provenance/p8e/shared/service/AffiliateService.kt
```diff
@@ -53,17 +58,58 @@ class AffiliateService(
...
+
+    /**
+     *  Return the signer to be used
+     *
+     *  [publicKey] publicKey used to validate
+     *  [signer] The specific signer used to sign contracts
+     */
+    fun getSigner(publicKey: PublicKey): SignerImpl {
+        val affiliateRecord = get(publicKey)
+        return if(affiliateRecord?.keyType == SMARTKEY) {
```
scirner22, line 73: one nice thing if this was a `when` instead of an `if/else` is that if we were to add a third type enum it would give compilation errors here saying it wasn't implemented in the `when`. As it currently is a third type that wasn't handled here would fall into the `else`.

os-client/src/main/kotlin/io/provenance/os/domain/inputstream/DIMEInputStream.kt
```diff
@@ -177,13 +183,13 @@ class DIMEInputStream(
...
-    fun getDecryptedPayload(keyPair: KeyPair): SignatureInputStream {
+    fun getDecryptedPayload(encryptionKeyRef: KeyRef, signer: SignerImpl): SignatureInputStream {
         // seek past the header
         pos += header.size
 
-        return ProvenanceDIME.getDEK(dime.audienceList, keyPair)
+        return ProvenanceDIME.getDEK(dime.audienceList, encryptionKeyRef)
             .let { ProvenanceDIME.decryptPayload(this, it) }
-            .verify(getFirstSignaturePublicKey(), getFirstSignature())
+            .verify(signer, getFirstSignaturePublicKey(), getFirstSignature())
```
scirner22, line 192: where is this verify function? I just wanted to see what it is doing now but can't find it.

build.gradle
scirner22: KK sounds good.

os-client/src/main/kotlin/io/provenance/os/domain/inputstream/DIMEInputStream.kt
rchaing-figure: https://github.com/provenance-io/p8e/blob/5eb9b7eb60fbb70537faa9466602104f290c77d4/p8e-util/src/main/kotlin/io/p8e/crypto/SignatureInputStream.kt#L56

p8e-api/src/main/kotlin/io/provenance/engine/service/MailboxService.kt
rchaing-figure: Yea, since we kept the encrypting the same, reverted.

os-client/src/main/kotlin/io/provenance/os/domain/inputstream/DIMEInputStream.kt
scirner22: thanks!

p8e-api-webservice/src/main/kotlin/io/provenance/p8e/webservice/config/AppConfig.kt
rchaing-figure: im going to clean this up, and just pass an uninstantiated instance of the API so that at least we know what the problem is if there are issues.

p8e-api/src/main/kotlin/io/provenance/engine/config/AppConfig.kt
rchaing-figure: im going to clean this up, and just pass an uninstantiated instance of the API so that at least we know what the problem is if there are issues.

p8e-api/src/main/kotlin/io/provenance/engine/grpc/observers/EnvelopeEventObserver.kt
rchaing-figure: yea, this might be a temp solution for now, until we start work on the clean envelope event observer ticket. (adding auth public key to the signing and encryption key proto).

p8e-api/src/main/kotlin/io/provenance/engine/grpc/v1/ObjectGrpc.kt
rchaing-figure: Yea, reverting back

p8e-api/src/main/kotlin/io/provenance/engine/batch/MailboxReaper.kt
rchaing-figure: The purpose of the signer for this particular call is to tell P8e which signer to init, actual verification doesn't happen until a little later below.

p8e-migration/sql/V69__Add_UUID_Key_To_Affiliate.sql
```diff
@@ -1,2 +1,2 @@
 ALTER TABLE affiliate ADD COLUMN signing_key_uuid UUID;
-ALTER TABLE p8e.affiliate ADD COLUMN key_provider_type TEXT NOT NULL DEFAULT 'DATABASE';
+ALTER TABLE affiliate ADD COLUMN key_provider_type TEXT NOT NULL DEFAULT 'DATABASE';
```
scirner22, line 2: I think this can't change now since it was already applied in figure test and now the file hash won't match.

rchaing-figure: what is the work around for this then? test doesn't know what the `p8e` schema is

p8e-shared/src/main/kotlin/io/provenance/p8e/shared/service/AffiliateService.kt
```diff
@@ -53,17 +59,54 @@ class AffiliateService(
...
+     *
+     *  [publicKey] publicKey used to validate
+     *  [signer] The specific signer used to sign contracts
+     */
+    fun getSigner(publicKey: PublicKey): SignerImpl {
+        val affiliateRecord = get(publicKey)
+        return when(affiliateRecord?.keyType) {
+            SMARTKEY -> signerFactory.getSigner(SignerFactoryParam.SmartKeyParam(affiliateRecord.signingKeyUuid.toString()))
+            DATABASE -> signerFactory.getSigner(SignerFactoryParam.PenParam(KeyPair(affiliateRecord.publicKey.value.toJavaPublicKey(), affiliateRecord.privateKey!!.toJavaPrivateKey())))
+            else -> throw UnsupportedOperationException("Unsupported signer type found ${affiliateRecord?.keyType?.name}.")
```
scirner22, line 77: You should be able to remove this case. Looks like it's only needed when affiliateRecord is null but in that case we can return an affiliate not found or something before the `when` expression.

p8e-migration/sql/V69__Add_UUID_Key_To_Affiliate.sql
scirner22: So this migration hasn't been applied in test? How does the branch code run without it?

rchaing-figure: misunderstood the timeline, this has already been applied to test, after removing of `p8e.` and is currently working fine.

scirner22: 👍

scirner22 (approved):

PR #25, opened 2021-04-21
Get public key from SmartKey and convert to BC

rchaing-figure:
Retrieve public key from SmartKey and convert it to a Java security public key so that it can be usable within P8e.

https://app.clubhouse.io/figure/story/89541/retrieve-public-keys-from-smartkey

Wanted to push this small addition out, before pushing other larger ones.

### DIFF
--- a/p8e-util/src/main/kotlin/io/p8e/crypto/SmartKeySigner.kt
+++ b/p8e-util/src/main/kotlin/io/p8e/crypto/SmartKeySigner.kt
@@ -3,6 +3,7 @@ package io.p8e.crypto
 import com.fortanix.sdkms.v1.ApiClient
 import com.fortanix.sdkms.v1.Configuration
 import com.fortanix.sdkms.v1.api.AuthenticationApi
+import com.fortanix.sdkms.v1.api.SecurityObjectsApi
 import com.fortanix.sdkms.v1.api.SignAndVerifyApi
 import com.fortanix.sdkms.v1.auth.ApiKeyAuth
 import com.fortanix.sdkms.v1.model.DigestAlgorithm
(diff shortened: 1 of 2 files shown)


p8e-util/src/main/kotlin/io/p8e/crypto/SmartKeySigner.kt
```diff
@@ -43,27 +71,39 @@ class SmartKeySigner(
...
+        val x509PublicKey = KeyFactory.getInstance("EC").generatePublic(X509EncodedKeySpec(smPublicKey))
+        val bcPublicKey = BCECPublicKey(x509PublicKey as ECPublicKey, BouncyCastlePQCProvider.CONFIGURATION)
+        return bcPublicKey.toHex().toJavaPublicKey()
+    }
+
     private fun verify(data: String, signature: Signature): Boolean = verify(data, signature)
 
-    private fun verify(data: ByteArray, signature: Signature): Boolean {
+    private fun verify(data: ByteArray, signature: ByteArray): Boolean {
+        //TODO: Investigate local verification: https://app.clubhouse.io/figure/story/90457/investigate-signature-verification-for-smartkey
```
rchaing-figure, line 100: Added a ticket to investigate local signature verification, tried to do it straight-up but the verification came back as false positive.

p8e-util/src/main/kotlin/io/p8e/crypto/SmartKeySigner.kt
```diff
@@ -43,27 +71,38 @@ class SmartKeySigner(
...
+        val x509PublicKey = KeyFactory.getInstance("EC").generatePublic(X509EncodedKeySpec(smPublicKey))
+        val bcPublicKey = BCECPublicKey(x509PublicKey as ECPublicKey, BouncyCastlePQCProvider.CONFIGURATION)
+        return bcPublicKey.toHex().toJavaPublicKey()
+    }
+
     private fun verify(data: String, signature: Signature): Boolean = verify(data, signature)
 
-    private fun verify(data: ByteArray, signature: Signature): Boolean {
+    private fun verify(data: ByteArray, signature: ByteArray): Boolean {
+        //TODO: Investigate local signature verification verification.
```
rchaing-figure, line 100: Added a ticket to investigate local signature verification, tried to do it straight-up but the verification came back as false positive.

https://app.clubhouse.io/figure/story/90457/investigate-signature-verification-for-smartkey

p8e-util/src/main/kotlin/io/p8e/crypto/SmartKeySigner.kt
```diff
@@ -11,20 +12,47 @@ import com.fortanix.sdkms.v1.model.VerifyRequest
...
-    appApiKey: String,
-    uuid: String
+    private val appApiKey: String,
+    private val keyUuid: String
 ): SignerImpl {
 
-    private var sObjUUID: String = uuid
-    private var apiKey: String = appApiKey
-
     init {
```
scirner22, line 53: This can be in a separate ticket, but given that this class now has many instances that could be signing simultaneously, I think we should make a bean for the actual api instance and have this class just hold a reference to the single client.

Likewise, if I remember correctly, the oauth token has some limited time duration so the singleton client class would need a background thread that updates the oauth token on a recurring basis.

scirner22 (approved):

p8e-util/src/main/kotlin/io/p8e/crypto/SmartKeySigner.kt
rchaing-figure: Good point, I'll create tickets for them.

piercetrey-figure (approved):

PR #26, opened 2021-04-22
Add affiliateService updates to support SmartKey

rchaing-figure:
Update methods in AffiliateService and Affiliate to better support with the SmartKey integration. 
Added overloading methods to support both Pen and SmartKey.

### DIFF
--- a/p8e-migration/sql/V66__Add_UUID_Key_To_Affiliate.sql
+++ b/p8e-migration/sql/V66__Add_UUID_Key_To_Affiliate.sql
@@ -1,0 +1,1 @@
+ALTER TABLE affiliate ADD COLUMN key_uuid UUID;
--- a/p8e-shared/src/main/kotlin/io/provenance/p8e/shared/domain/Affiliate.kt
+++ b/p8e-shared/src/main/kotlin/io/provenance/p8e/shared/domain/Affiliate.kt
@@ -2,6 +2,8 @@ package io.provenance.p8e.shared.domain
 
 import io.p8e.proto.Affiliate.AffiliateWhitelist
 import io.p8e.util.toHex
+import io.p8e.util.toJavaPublicKey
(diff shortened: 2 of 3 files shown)


p8e-migration/sql/V66__Add_UUID_Key_To_Affiliate.sql
```diff
@@ -0,0 +1 @@
+ALTER TABLE affiliate ADD COLUMN key_uuid UUID;
```
rchaing-figure, line 1: Thought about making this a PK, might not be necessary, it adds more risk and effort as we have 6 tables that are FK to the public key.

rchaing-figure:
Just adding method support for the SmartKey integration (in another branch) tested that existing functionality still works as is.

piercetrey-figure (approved):

p8e-migration/sql/V66__Add_UUID_Key_To_Affiliate.sql
scirner22: I totally forgot about signing and encryption keys being able to be different now. Shouldn't we add a signing_key_uuid and encryption_key_uuid here so they can be referenced independently?

rchaing-figure: Right, was debating that I wanted to work on signing keys (the tougher of the 2 keys) first, get it working E2E, then start on encryption, think if I tried to tackle both of the keys at the same time, it might be very regression prone and confusing (like how the diff key PR was). When I get around to encryption I'll just PR a branch against the feature.

scirner22: That's a good plan. I just wanted to mention it as it was on my mind.

scirner22 (approved):

PR #27, opened 2021-05-03
fix event stream duplication

piercetrey-figure:
## Description

if there were network or other issues that caused an event stream restart, and those network issues also caused errors in the streamHistory logic, then the event stream websocket could end up connected multiple times, duplicating broadcasted events

### DIFF
--- a/p8e-shared/src/main/kotlin/io/provenance/p8e/shared/util/MDC.kt
+++ b/p8e-shared/src/main/kotlin/io/provenance/p8e/shared/util/MDC.kt
@@ -62,9 +62,9 @@ object P8eMDC {
     }
 }
 
-data class TransactionHashes(val hashes: List<String>)
+data class TransactionHashes(val hashes: Collection<String>)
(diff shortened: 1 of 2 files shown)


p8e-shared/src/main/kotlin/io/provenance/p8e/shared/util/MDC.kt
```diff
@@ -62,9 +62,9 @@ object P8eMDC {
     }
 }
 
-data class TransactionHashes(val hashes: List<String>)
+data class TransactionHashes(val hashes: Collection<String>)
 fun String.toTransactionHashes() = TransactionHashes(listOf(this))
-fun List<String>.toTransactionHashes() = TransactionHashes(this)
+fun List<String>.toTransactionHashes() = TransactionHashes(this.toSet())
```
piercetrey-figure, line 67: this was just to clean up the mdc log when a tx had a bunch of envelopes in it

scirner22 (approved):

PR #30, opened 2021-05-05
prevent mailing error to other members on same p8e instance

piercetrey-figure:
- the EnvelopeRecords found by execution uuid will handle the error state for these parties, any other parties can still be reached via mailbox for now


p8e-api/src/main/kotlin/io/provenance/engine/service/TransactionStatusService.kt
```diff
@@ -33,21 +33,25 @@ class TransactionStatusService(
...
-                    envelope.scope.publicKey.toJavaPublicKey(),
-                    envelope.data.input,
-                    it
-                )
+                    // Ship error info to an relevant additional parties
+                    mailboxService.error(
+                        envelope.scope.publicKey.toJavaPublicKey(),
+                        envelope.data.input,
+                        it,
+                        *sameInstanceKeys
```
piercetrey-figure, line 52: mailbox service filters out keys in this list so they won't be mailed to

scirner22 (approved):

PR #32, opened 2021-05-06
Add support for auth public key

rchaing-figure:
Add the ability to auth against p8e-api with a auth public key. 

Motivation behind this change is to restrict the usage of private keys within the code base especially when adding a key management system where the private key is not easily accessible. Also, existing keys already serves a purpose throughout the entire system where adding an auth key relieves the security risks of P8e having to manage the private key and is now in the complete possession of the user when auth-ing against p8e-api.

### DIFF
--- a/p8e-api-webservice/src/main/kotlin/io/provenance/p8e/webservice/domain/ApiAffiliate.kt
+++ b/p8e-api-webservice/src/main/kotlin/io/provenance/p8e/webservice/domain/ApiAffiliate.kt
@@ -43,9 +43,10 @@ fun AffiliateShareRecord.toApi(): ApiAffiliateShare =
         created
     )
 
-data class RegisterAffiliateKey(val signingPrivateKey: String?, val encryptionPrivateKey: String?, val useSigningKeyForEncryption: Boolean, val indexName: String, val alias: String?) {
+data class RegisterAffiliateKey(val signingPrivateKey: String?, val encryptionPrivateKey: String?, val authKeyPair: String?, val useSigningKeyForEncryption: Boolean, val indexName: String, val alias: String?) {
     val signingKeyPair: KeyPair = signingPrivateKey.toOrGenerateKeyPair()
     val encryptionKeyPair: KeyPair = signingKeyPair.takeIf { useSigningKeyForEncryption }.or { encryptionPrivateKey.toOrGenerateKeyPair() }
+    val authPublicKey: PublicKey = authKeyPair.toOrGenerateKeyPair().public
 
     private fun String?.toOrGenerateKeyPair() = this?.takeIf { it.isNotBlank() }?.toJavaPrivateKey()?.let {
         KeyPair(it.computePublicKey(), it)
--- a/p8e-api-webservice/src/main/kotlin/io/provenance/p8e/webservice/domain/ApiAffiliate.kt
+++ b/p8e-api-webservice/src/main/kotlin/io/provenance/p8e/webservice/domain/ApiAffiliate.kt
@@ -43,10 +43,10 @@ fun AffiliateShareRecord.toApi(): ApiAffiliateShare =
         created
     )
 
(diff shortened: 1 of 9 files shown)


p8e-proto-internal/src/main/proto/p8e/affiliate.proto
```diff
@@ -18,6 +18,7 @@ service AffiliateService {
 message AffiliateRegisterRequest {
     PrivateKey signing_private_key = 1;
     PrivateKey encryption_private_key = 2;
+    PublicKey auth_public_key = 3;
```
rchaing-figure, line 21: Don't think we are using the register method in AffiliateGrpc for anything, but add this just in case we decide to, we don't forget about this.

p8e-api-webservice/src/main/kotlin/io/provenance/p8e/webservice/domain/ApiAffiliate.kt
```diff
@@ -43,9 +43,10 @@ fun AffiliateShareRecord.toApi(): ApiAffiliateShare =
         created
     )
 
-data class RegisterAffiliateKey(val signingPrivateKey: String?, val encryptionPrivateKey: String?, val useSigningKeyForEncryption: Boolean, val indexName: String, val alias: String?) {
+data class RegisterAffiliateKey(val signingPrivateKey: String?, val encryptionPrivateKey: String?, val authKeyPair: String?, val useSigningKeyForEncryption: Boolean, val indexName: String, val alias: String?) {
```
piercetrey-figure, line 46: Looks like this is actually the private key in this request, right? (i.e. the name authKeyPair might be misleading)

p8e-api-webservice/src/main/kotlin/io/provenance/p8e/webservice/domain/ApiAffiliate.kt
```diff
@@ -43,9 +43,10 @@ fun AffiliateShareRecord.toApi(): ApiAffiliateShare =
...
     val signingKeyPair: KeyPair = signingPrivateKey.toOrGenerateKeyPair()
     val encryptionKeyPair: KeyPair = signingKeyPair.takeIf { useSigningKeyForEncryption }.or { encryptionPrivateKey.toOrGenerateKeyPair() }
+    val authPublicKey: PublicKey = authKeyPair.toOrGenerateKeyPair().public
```
piercetrey-figure, line 49: within this class, I think we might need to actually keep the whole key pair, as the response will include the private key (only time we send the private key back, so the user has it in the case that it was generated)

piercetrey-figure (approved):
Just the naming and private key comments, otherwise looks good. I can make the corresponding UI changes to supply/display this key via the modal when we get closer to merging the signer stuff in

p8e-migration/sql/V67__Add_Auth_Public_Key_to_Affiliate.sql
```diff
@@ -0,0 +1 @@
+ALTER TABLE affiliate ADD COLUMN auth_public_key TEXT;
```
piercetrey-figure, line 1: @rchaing-figure should we go ahead and have this migration just copy over the signing_public_key value for existing affiliates and make this column not nullable?

rchaing-figure: Wasn't ready to do this yet until I get the Encryption Key all setup as well, then see how this all looks and see which key would make more sense.

p8e-api-webservice/src/main/kotlin/io/provenance/p8e/webservice/domain/ApiAffiliate.kt
rchaing-figure: Got it, wasn't clear of how this would look from the UI perspective and if was strict on which key should be used/set. I'll make the change.

PR #38, opened 2021-05-17
SmartKey Signing P8e Integration

rchaing-figure:
Initial integration of SmartKey into P8e.

- Implicit signer that chooses between the default Java security signer or SmartKey
- Signer is selected based on private key availability in the P8e affiliate database.

** Assumption Mailbox Client will go away.


p8e-util/src/main/kotlin/io/p8e/crypto/SmartKeySigner.kt
```diff
@@ -45,64 +48,119 @@ import java.security.spec.X509EncodedKeySpec
...
+            /**
+             * Large data dump from the InputStream drops in here, we only care about
+             * the last piece of data to verify against. If we continue to update the
+             * signature object with every single data that pass through the verification
+             * returns a false negative.
+             *
+             * Signature.getInstance is a 1:1 - meaning new instance for each piece of data.
+             */
+
+            cachedData.add(data.copyOfRange(off, off+len))
```
rchaing-figure, line 105: Not a big fan of this, looking into ways to improve this, ideas appreciated. Essentially, when we push to OS, and then fetch the data for verification, the data that we signed for gets appended to the end of the input stream, we don't care for anything before and at the same time we don't want to mess up our local signature object during data update (Hence, L#111).

os-client/src/main/kotlin/io/provenance/os/client/OsClient.kt
```diff
@@ -173,7 +174,7 @@ open class OsClient(uri: URI) {
             dime.dime,
             dime.encryptedPayload,
             uuid = uuid,
-            metadata = metadata + (SIGNATURE_PUBLIC_KEY_FIELD_NAME to CertificateUtil.publicKeyToPem(signingKeyPair.public)),
+            metadata = metadata + (SIGNATURE_PUBLIC_KEY_FIELD_NAME to CertificateUtil.publicKeyToPem(signerPublicKey)),
```
piercetrey-figure, line 177: probably can replace the CertificateUtil.publicKeyToPem(signerPublicKey) with `signingPublicKey` from above (to avoid computing the same value twice)

p8e-api/src/main/kotlin/io/provenance/engine/grpc/v1/IndexGrpc.kt
```diff
@@ -204,7 +205,8 @@ class IndexGrpc(
                         definitionService.get(
                             keyPair,
```
piercetrey-figure, line 206: this appears to be the signing key pair, but I think DefinitionService is expecting the encryption key pair, right? I'm wondering if this is just slightly broken (not that it matters, as the key pair will go away once the encryption abstraction is in place, right?)

p8e-common/src/main/kotlin/io/p8e/definition/DefinitionService.kt
```diff
@@ -86,38 +91,44 @@ class DefinitionService(
     fun loadProto(
         keyPair: KeyPair,
         location: Location,
+        signer: SignerImpl,
         signaturePublicKey: PublicKey? = null
```
piercetrey-figure, line 95: seems like maybe we could avoid passing the signaturePublicKey through all these functions since it can be fetched from the SignerImpl, right?

p8e-api/src/main/kotlin/io/provenance/engine/index/ProtoIndexer.kt
```diff
@@ -41,21 +44,25 @@ class ProtoIndexer(
                             .first { keyPairs.containsKey(it.encryptionPublicKey.toHex()) }
                             .let { keyPairs.getValue(it.encryptionPublicKey.toHex()) }
 
+                        // Need a reference of the signer that is used to verify signatures.
+                        val signer = transaction { affiliateService.getSigner(keyPair.public) }
```
piercetrey-figure, line 48: looks like you were able to get rid of the transaction around the getSigner call on line 78 of ObjectGrpc (looks like there isn't any db stuff in that function ?), can you nix it everywhere else as well?

p8e-util/src/main/kotlin/io/p8e/crypto/SignatureInputStream.kt
```diff
@@ -1,21 +1,64 @@
...
     private val inputStream: InputStream,
-    private val signature: Signature
+    private val signer: SignerImpl,
+    private val signatureBytes: ByteArray? = null
 ): InputStream() {
     override fun read(): Int {
         val res = inputStream.read()
         if (res != -1) {
-            signature.update(res.toByte())
+            signer.update(inputStream.readAllBytes())
```
piercetrey-figure, line 14: I assume this starts from the beginning of the stream (i.e. the read call above didn't advance one byte into the stream or something like that?)

rchaing-figure: don't think this read method gets invoked, the other read with the off and len does, didn't spend anytime time looking if the other 2 reads are used for something else

p8e-api/src/main/kotlin/io/provenance/engine/index/ProtoIndexer.kt
rchaing-figure: ill take care of it

piercetrey-figure (approved):
A couple of questions and ideas, but overall I think it is making sense

p8e-common/src/main/kotlin/io/p8e/definition/DefinitionService.kt
rchaing-figure: was hesitant to do this, as some method calls don't pass a public key on purpose thats why i passed at the calling level

p8e-api/src/main/kotlin/io/provenance/engine/grpc/v1/IndexGrpc.kt
rchaing-figure: good 👁️ , unsure if you need laser

os-client/src/main/kotlin/io/provenance/os/client/OsClient.kt
rchaing-figure: ill fix this up

p8e-util/src/main/kotlin/io/p8e/crypto/SignatureInputStream.kt
rchaing-figure: crap .. yea, i have to revert that too

p8e-api/src/main/kotlin/io/provenance/engine/index/ProtoIndexer.kt
rchaing-figure: actually just saw line 78 of object grpc, its wrapped with a transaction one line above

p8e-shared/src/main/kotlin/io/provenance/p8e/shared/service/AffiliateService.kt
```diff
@@ -46,6 +50,26 @@ class AffiliateService(
...
+     *  [signer] The specific signer used to sign contracts
+     */
+    fun getSigner(publicKey: PublicKey): SignerImpl {
+        val affiliateRecord = get(publicKey)
+        return if(affiliateRecord?.privateKey == null) {
+            signerFactory.getSigner(SignerFactoryParam.SmartKeyParam(affiliateRecord?.keyUuid.toString()))
+        } else {
+            signerFactory.getSigner(
+                SignerFactoryParam.PenParam(
+                    KeyPair(affiliateRecord.publicKey.value.toJavaPublicKey(), affiliateRecord.privateKey?.toJavaPrivateKey())
```
scirner22, line 69: Can the `privateKey?` become `privateKey`? I think kotlin should have the information to infer that it's non-nullable in this expression based on the `affiliateRecord?.privateKey == null` statement.

scirner22 (approved):

p8e-shared/src/main/kotlin/io/provenance/p8e/shared/service/AffiliateService.kt
rchaing-figure: Looks like it still complains about it, telling me to `!!` it.

p8e-api/src/main/kotlin/io/provenance/engine/index/ProtoIndexer.kt
piercetrey-figure: oh yeah, missed that... also just looked at the getSigner function, and I guess it does fetch the affiliate to determine which signer to use, so :nothingtodohere:

PR #43, opened 2021-05-21
Ch88576/webservice smartkey updates

piercetrey-figure:
- Working adding affiliates using either database or smart key provider 
  - Both keys have the APPMANAGEABLE operation set so we can update them programmatically (i.e. change name, etc.)
  - Signing key can sign/verify only
  - Encryption key has the AGREEKEY operation so it can be used for the diffie-hellman key agreement used for the encryption process w/ object-store 🤞 
- Working edit of Affiliate alias being reflected in SmartKey
- If there is an error during smartKey affiliate creation (either generating a key, or something database related), any generated keys are deleted, so as to not leave orphaned keys hanging out in SmartKey

### DIFF
--- a/p8e-migration/sql/V70__Adjust_Encryption_Key_Columns.sql
+++ b/p8e-migration/sql/V70__Adjust_Encryption_Key_Columns.sql
@@ -1,0 +1,3 @@
+ALTER TABLE p8e.affiliate
+    ADD COLUMN encryption_key_uuid UUID,
+    ALTER COLUMN encryption_private_key DROP NOT NULL;
--- a/p8e-migration/sql/V70__Adjust_Encryption_Key_Columns.sql
+++ b/p8e-migration/sql/V70__Adjust_Encryption_Key_Columns.sql
@@ -1,3 +1,3 @@
-ALTER TABLE p8e.affiliate
+ALTER TABLE affiliate
     ADD COLUMN encryption_key_uuid UUID,
     ALTER COLUMN encryption_private_key DROP NOT NULL;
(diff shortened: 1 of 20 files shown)


p8e-migration/sql/V70__Adjust_Encryption_Key_Columns.sql
```diff
@@ -0,0 +1,3 @@
+ALTER TABLE p8e.affiliate
```
rchaing-figure, line 1: probably want to leave out the schema name in case we decide to change it in the future.

p8e-shared/src/main/kotlin/io/provenance/p8e/shared/service/AffiliateService.kt
```diff
@@ -296,6 +296,11 @@ class AffiliateService(
                         require (response.isAcknowledged) { "ES index creation of $it was not successful" }
                     }
                 }
+
+                if (jwt != null && identityUuid != null) {
+                    keystoneService.registerKey(jwt, signingPublicKey.publicKey, ECUtils.LEGACY_DIME_CURVE, KeystoneKeyUsage.CONTRACT)
```
rchaing-figure, line 301: do we still need keystone for anything now? this might be moving to smartkey soon too.

p8e-migration/sql/V70__Adjust_Encryption_Key_Columns.sql
piercetrey-figure: oh, oops, put that in to make the intelliJ sql editor happy, I'll remove it

p8e-shared/src/main/kotlin/io/provenance/p8e/shared/service/AffiliateService.kt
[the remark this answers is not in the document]
piercetrey-figure: I just copied this block from the other save function, this gets removed in the branch that uses the new object store (since that is where the oauth authentication removal is as well), so I just copied it all to keep it consistent until it all gets merged together

scirner22 (approved):

rchaing-figure (approved):

PR #47, opened 2021-06-02
SmartKey Encryption Integration

rchaing-figure:
- Added Encryption Key support for SmartKey.


p8e-api/src/main/kotlin/io/provenance/engine/grpc/v1/IndexGrpc.kt
```diff
@@ -169,8 +169,8 @@ class IndexGrpc(
     ) {
         P8eMDC.set(publicKey(), clear = true)
 
-        val keyPair = transaction { affiliateService.getSigningKeyPair(publicKey()) }
         val signer = transaction { affiliateService.getSigner(publicKey()) }
+        val encryptionKeyRef = transaction { affiliateService.getEncryptionKeyRef(publicKey()) }
```
scirner22, line 173: So this changed from signing to encryption key. Was this previously a bug we never hit yet?

rchaing-figure: All affiliate keys (both signing and encryption) are the same in Test and Prod I believe, so that's why it is a non-issue, but if we start having people use different signing and encryption then there would probably be an issue.

scirner22 (approved):

PR #51, opened 2021-06-08
get client ip in grpc interceptor and log on EnvelopeEventObserver connect

piercetrey-figure:
- Might be helpful in tracking down weird connection issues (i.e. determining if two different services are trying to watch with the same key/contract combo)

### DIFF
--- a/p8e-api/src/main/kotlin/io/provenance/engine/grpc/observers/EnvelopeEventObserver.kt
+++ b/p8e-api/src/main/kotlin/io/provenance/engine/grpc/observers/EnvelopeEventObserver.kt
@@ -1,6 +1,7 @@
 package io.provenance.engine.grpc.observers
 
 import io.grpc.stub.StreamObserver
+import io.p8e.grpc.clientIp
 import io.p8e.grpc.observers.CompleteState
 import io.p8e.grpc.observers.EndState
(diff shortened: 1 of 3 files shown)


p8e-api/src/main/kotlin/io/provenance/engine/grpc/observers/EnvelopeEventObserver.kt
```diff
@@ -168,6 +169,8 @@ class EnvelopeEventObserver(
     private fun connect(value: EnvelopeEvent): QueueingStreamObserverSender<EnvelopeEvent> {
         val publicKey = if(value.publicKey.hasEncryptionPublicKey()) value.publicKey.encryptionPublicKey.toPublicKey() else value.publicKey.signingPublicKey.toPublicKey()
 
+        logger().info("Connecting [publicKey = ${publicKey.toHex()}, classname = ${value.classname}, action = ${value.action}, ip = ${clientIp()}]")
```
scirner22, line 172: I think it makes sense to have this at the debug level, that would require us finding a way to control the logging level outside of this repo and with p8e-deployment or the figure-deployment-controller directly.

scirner22 (approved):

PR #53, opened 2021-06-13
Re-Auth SmartKey's token

rchaing-figure:
- SmartKey API token expires every 10 minutes, background thread will refresh at every 5 minutes.


p8e-api-webservice/src/main/kotlin/io/provenance/p8e/webservice/config/AppConfig.kt
```diff
@@ -113,14 +113,16 @@ class AppConfig : WebMvcConfigurer {
...
 
+    @Bean
+    fun authenticationApi(smartKeyApiClient: ApiClient): AuthenticationApi = AuthenticationApi(smartKeyApiClient)
+
     @Bean
     fun securityObjectsApi(smartKeyApiClient: ApiClient): SecurityObjectsApi = SecurityObjectsApi(smartKeyApiClient)
 
     @Bean
-    fun smartKeySigner(signAndVerifyApi: SignAndVerifyApi, securityObjectsApi: SecurityObjectsApi): SmartKeySigner = SmartKeySigner(signAndVerifyApi, securityObjectsApi)
+    fun smartKeySigner(signAndVerifyApi: SignAndVerifyApi, securityObjectsApi: SecurityObjectsApi, authenticationApi: AuthenticationApi): SmartKeySigner
```
rchaing-figure, line 123: @piercetrey-figure  Is a different SmartKeySigner instance created at WebService time? If so I'd imagine it's own instance would be able to refresh it's own token.

p8e-util/src/main/kotlin/io/p8e/crypto/SmartKeySigner.kt
```diff
@@ -44,10 +45,14 @@ import java.security.Signature
...
     private val signAndVerifyApi: SignAndVerifyApi,
     private val securityObjectsApi: SecurityObjectsApi,
+    private val authenticationApi: AuthenticationApi
 ): SignerImpl {
 
     init {
         Security.addProvider(BouncyCastleProvider())
+
+        // Kick background thread to refresh SmartKey's auth api.
+        SmartKeyAuthRefresh(authenticationApi).refreshSmartKeyAuthToken()
```
rchaing-figure, line 55: Refreshing token at the signer, cause before we can encrypt we would need to sign the data.

p8e-api-webservice/src/main/kotlin/io/provenance/p8e/webservice/config/AppConfig.kt
piercetrey-figure: the WebService isn't using the signer directly, just the securityObjectsApi (this bean is to satisfy the SignerFactory being injected into the AffiliateService I think), but it should be able to do its own token refreshing, so either it can call the SmartKeyAuthRefresh thing in the KeyManagementService, or maybe the SmartKeySigner is getting intialized anyways... though probably not as the factory would have to be called, right?

p8e-util/src/main/kotlin/io/p8e/crypto/SmartKeyAuthRefresh.kt
```diff
@@ -0,0 +1,34 @@
+package io.p8e.crypto
+
+import com.fortanix.sdkms.v1.api.AuthenticationApi
+import com.google.common.util.concurrent.ThreadFactoryBuilder
+import java.util.concurrent.Executors
+import java.util.concurrent.TimeUnit.MINUTES
+import kotlin.concurrent.thread
+
+class SmartKeyAuthRefresh(
```
piercetrey-figure, line 9: would it make more sense to just have this in p8e-shared and registered as an @Component, with an @Scheduled annotation on the refreshSmartKeyAuthToken to schedule it? (maybe with a guard to not do anything if there is no smartKey api key... which I guess we need to make nullable...) Then the SmartKeySigner wouldn't have to be responsible for kicking this off...

p8e-api-webservice/src/main/kotlin/io/provenance/p8e/webservice/config/AppConfig.kt
rchaing-figure: cool, so the AuthRefresh is being setup during SmartKeySigner's init, though im guessing p8e-api doesn't have visibility to the KeyManagementService? Id rather have the AuthRefresh happen at one place.

piercetrey-figure: Yeah, I don't think p8e-api is using the KeyManagementService, though it could if needed. I think it is probably fine where it is, all I was saying is it could maybe be an entirely separate concern (separate from key management, or signing, just something that keeps the token alive for whatever might need it)

p8e-util/src/main/kotlin/io/p8e/crypto/SmartKeyAuthRefresh.kt
rchaing-figure: ill make the change, seems like the better solution.

piercetrey-figure (approved):

p8e-shared/src/main/kotlin/io/provenance/p8e/shared/util/SmartKeyAuthRefresh.kt
```diff
@@ -0,0 +1,29 @@
...
+) {
+
+    private val log = logger()
+
+    companion object {
+        const val REFRESH_RATE: Long = 300_000 // refresh at a rate of every 5 minutes.
+    }
+
+    /**
+     * Refresh the Auth Token every 5 Minutes
```
scirner22, line 20: Does refresh use the existing api key to fetch a new one or only based off of the api secret key?

rchaing-figure: It extends the current session. According to Smart Key:

`Refreshes an existing session, thus extending the expiration time of the existing bearer authentication token`

Ill reword the comment

scirner22 (approved):

PR #57, opened 2021-06-15
Feature/publish maven central

scirner22:
## Description

Now that we have approval to publish to Maven Central, we should make use of that. What we saw in the past was third parties struggled to get the GITHUB_API keys properly setup in order to make use of the github repository.


.github/workflows/release-build-publish.yml
```diff
@@ -55,11 +55,23 @@ jobs:
...
+      - name: Publish to Maven Central
+        run: |
+          ./gradlew publish \
+            -Pversion=$VERSION \
+            -Psigning.keyId=B7D30ABE -Psigning.password="${{ secrets.OSSRH_GPG_SECRET_KEY_PASSWORD }}" -Psigning.secretKeyRingFile=$GITHUB_WORKSPACE/release.gpg \
+            --info
         env:
-          GITHUB_ACTOR: ${{ github.actor }}
-          GITHUB_TOKEN: ${{ secrets.GITHUB_TOKEN }}
+          OSSRH_USERNAME: ${{ secrets.OSSRH_USERNAME }}
```
piercetrey-figure, line 73: It appears these are all repository secrets currently, are you planning on making them organization-level secrets now that you have them all figured out so we can publish from other provenance-io repositories?

scirner22: Yes, the next time we add something that needs them I think that would make more sense.

scirner22:
One outstanding thing here -- this pushes the release jars to the staging repo in sonatype. I believe I need to wait for the sonatype support to link the auto sync to the actual maven central repo.

rchaing-figure (approved):

PR #61, opened 2021-06-24
build docker multi-architecture

piercetrey-figure:
- make things work for m1
- build arm64 AND amd64 images
- tried this out on the p8e-ui repo and it seems to work

### DIFF
--- a/.github/workflows/release-build-publish.yml
+++ b/.github/workflows/release-build-publish.yml
@@ -119,6 +119,22 @@ jobs:
 
           echo "VERSION=$VERSION" >> $GITHUB_ENV
 
+      - name: Setup QEMU
+        id: qemu
+        uses: docker/setup-qemu-action@v1
+        with:
+          image: tonistiigi/binfmt:latest
+          platforms: all
+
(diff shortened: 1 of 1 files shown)


.github/workflows/release-build-publish.yml
```diff
@@ -119,36 +119,50 @@ jobs:
...
+          platforms: all
+
+      - name: Set up Docker Buildx
+        id: buildx
+        uses: docker/setup-buildx-action@v1
+        with:
+          install: true
+
+      - name: Available platforms
+        run: echo ${{ steps.qemu.outputs.platforms }}
```
rchaing-figure, line 149: only used to print info?

rchaing-figure (approved):

scirner22 (approved):

.github/workflows/release-build-publish.yml
piercetrey-figure: yeah, just for easy help debugging if we ever need to build to additional platforms. I copied that from the provenance repo and saw it in some examples, so just decided to keep it

PR #63, opened 2021-06-29
Fix/expand pre execution error

scirner22:
## Description

A big problem with the `PreExecutionError` is that it comes before anything was persisted to the backend. This means that it's harder to debug from the server side. As a client, this should give them enough relevant information to help with debugging.


rchaing-figure (approved):

os-client/src/main/kotlin/io/provenance/os/client/OsClient.kt
```diff
@@ -86,18 +86,6 @@ open class OsClient(
             }
     }
 
-    fun get(uri: String, publicKey: PublicKey): DIMEInputStream {
```
piercetrey-figure, line 89: was this just no longer valid with the new object store or something?

piercetrey-figure (approved):

os-client/src/main/kotlin/io/provenance/os/client/OsClient.kt
scirner22: 🤷  I noticed nothing in p8e is using it right now. I'm not sure if it was used in the past or not.

PR #64, opened 2021-06-29
Add data access

wbaker-figure:
## Description

This PR adds Data access to the contract scope after memorialization. Files to Check:
IndexHandler.kt
DataAccessChaincodeService.kt


p8e-api/src/main/kotlin/io/provenance/engine/event/IndexHandler.kt
```diff
@@ -114,6 +123,36 @@ class IndexHandler(
                                     envelope.data.result.contract.recitalsList.map { it.signer.encryptionPublicKey.toPublicKey().toSha512Hex() }
                                 ), RequestOptions.DEFAULT
                             )
+                            // If the env is the invoker, create the data access message and put into a job.
+                            if(envelope.isInvoker == true) {
```
scirner22, line 127: We could save some time too by adding `&& envelope.data.input.affiliateShareLists.isNotEmpty()`.

p8e-api/src/main/kotlin/io/provenance/engine/event/IndexHandler.kt
```diff
@@ -114,6 +123,36 @@ class IndexHandler(
...
+                                                affiliateService.getAddress(it, chaincodeProperties.mainNet)
+                                            }
+                                        })
+                                        .setScopeId(
+                                            envelope.data.input.ref.scopeUuid.value.toUuidProv().toAddress(
+                                                PROV_METADATA_PREFIX_SCOPE_ADDR
+                                            ).toByteString()
+                                        )
+                                        .setPublicKey(envelope.data.input.contract.invoker.encryptionPublicKey)
+                                        .build().takeIf { envelope.data.input.affiliateSharesList.isNotEmpty() }
```
scirner22, line 149: I think it would be more idiomatic kotlin to remote the `if` below and instead add

```kotlin
?.let { dataAccessService.addDataAccess(it) }
```

p8e-api/src/main/kotlin/io/provenance/engine/service/ChaincodeInvokeService.kt
```diff
@@ -40,7 +43,7 @@ data class ContractRequestWrapper(
     val executionUuid: UUID,
     val request: MsgP8eMemorializeContractRequest,
     val future: CompletableFuture<ContractTxResult>,
-)
+    )
```
scirner22, line 46: Is your IDE automatically making these format changes?

p8e-api/src/main/kotlin/io/provenance/engine/service/ChaincodeInvokeService.kt
```diff
@@ -229,7 +230,8 @@ class ChaincodeInvokeService(
                         retryable -> {
                             handleBatchRetry(errorMessage)
                         }
-                        else -> { // fail the whole batch
```
scirner22, line 232: same here as above

p8e-api/src/main/kotlin/io/provenance/engine/service/ChaincodeInvokeService.kt
```diff
@@ -425,8 +425,8 @@ class ChaincodeInvokeService(
         } else {
             log.info("skipping gasMultiplier due to daily limit")
         }
-
-        provenanceGrpc.batchTx(body, accountNumber, sequenceNumber, estimate)
+        // TODO this is the hardcoded way to force the update through
```
scirner22, line 428: seems like this can probably be cleaned up now'?

p8e-api/src/main/kotlin/io/provenance/engine/service/ChaincodeInvokeService.kt
```diff
@@ -10,21 +10,24 @@ import io.p8e.proto.ContractSpecs.ContractSpec
 import io.p8e.proto.Contracts
 import io.p8e.util.*
 import io.provenance.engine.config.ChaincodeProperties
+import io.provenance.engine.crypto.Account
```
scirner22, line 13: You can probably revert this whole file actually. I think it was a remnant of our early testing.

p8e-api/src/main/kotlin/io/provenance/engine/service/ChaincodeInvokeService.kt
wbaker-figure: Uhhhhh yeah. Totally. For sure. 😅
I assume I did some odd stuff during merge of os-locator or when I was removing some of the code I had in there for initial testing. Will make sure it doesn't happen again.

p8e-api/src/main/kotlin/io/provenance/engine/event/IndexHandler.kt
```diff
@@ -114,6 +123,33 @@ class IndexHandler(
                                     envelope.data.result.contract.recitalsList.map { it.signer.encryptionPublicKey.toPublicKey().toSha512Hex() }
                                 ), RequestOptions.DEFAULT
                             )
+                            // If the env is the invoker, create the data access message and put into a job.
+                            if(envelope.isInvoker == true && envelope.data.input.affiliateSharesList.isNotEmpty()) {
+                                val envelopeDataAccess = envelope.data.input.affiliateSharesList.map {
```
piercetrey-figure, line 128: This function is already kind of massive, I think it might be nice if this data access block got moved into a separate (private) function with some lovely name in this class (whether just the bit inside of the if, or a function that contains the if statement as well, and otherwise is a no-op), just to make it a bit easier to parse what is happening when reading

p8e-api/src/main/kotlin/io/provenance/engine/service/DataAccessChaincodeService.kt
```diff
@@ -0,0 +1,131 @@
...
+        val affiliateAccount = provenanceGrpcService.accountInfo(affiliateAddress)
+
+        val estimate = provenanceGrpcService.estimateTx(
+                msgAddScopeDataAccessRequest.toTxBody(),
+                affiliateAccount.accountNumber,
+                affiliateAccount.sequence
+            )
+
+        // perform and wait for hash transfer to complete if account has less than 10 hash
+        if (provenanceGrpcService.getAccountCoins(affiliateAddress)[0].amount.toLong() / 1000000000 < 10) {
```
piercetrey-figure, line 57: Do we know if hash would always be the first coin on an account? Isn't it possible that there could be other coins at some point? Just wondering if we should do a find {} on the returned list to look it up by name or something?

piercetrey-figure: also, I think it might be nice to have this `1000000000` nhash -> hash factor in a constant that can be used here and down below for the transfer (i.e. `10 * HASH_TO_NHASH` or something like that), the 0's are hard to read 😄

piercetrey-figure (approved):
Looking good, nice job on working through figuring out who the heck is supposed to sign stuff 😆

p8e-api/src/main/kotlin/io/provenance/engine/event/IndexHandler.kt
```diff
@@ -114,6 +123,33 @@ class IndexHandler(
...
+                                    affiliateService.getAddress(
+                                        it.toPublicKey(), chaincodeProperties.mainNet
```
rchaing-figure, line 130: minor nit-picking after working with a bunch of keys, when following code you can get lost in the `it` when there is a bunch of nested `it`s ... so you can do `{ affiliateSharePublicKey -> affilliateSharePublicKey.toPublicKey() .. ` same thing for the ones below.

p8e-api/src/main/kotlin/io/provenance/engine/service/DataAccessChaincodeService.kt
```diff
@@ -0,0 +1,131 @@
...
+        val msgAddScopeDataAccessRequest = MsgAddScopeDataAccessRequest.newBuilder().addAllDataAccess(payload.msgAddScopeDataAccessRequest.dataAccessList)
+            .addAllSigners(payload.msgAddScopeDataAccessRequest.signersList)
+            .setScopeId(payload.msgAddScopeDataAccessRequest.scopeId)
+            .build()
+        val publicKey = payload.msgAddScopeDataAccessRequest.publicKey.toPublicKey()
+
+        val affiliate =
+            transaction { affiliateService.get(publicKey) }.orThrowNotFound("Affiliate with public key ${publicKey.toHex()} not found")
+        val affiliateKeyPair =
+            KeyPair(affiliate.publicKey.value.toJavaPublicKey(), affiliate.privateKey.toJavaPrivateKey())
```
rchaing-figure, line 44: We may need to add SmartKey support here as well ... probably setup data under KeyRef data class (hopefully this branch goes out before mine so you don't have to worry about it) :) 

Essentially instead of a KeyPair, we have KeyRef that contains key information such as public key, maybe private key?, UUID (SmartKey way of identifying a private key stored on their system).

p8e-api/src/main/kotlin/io/provenance/engine/event/IndexHandler.kt
```diff
@@ -169,4 +182,35 @@ class IndexHandler(
         request.opType(OpType.INDEX)
         return request
     }
+
+    private fun updateDataAccess(envelope: EnvelopeRecord) {
+        val envelopeDataAccess = envelope.data.input.affiliateSharesList.map {
+            affiliateSharePublicKey ->
+            affiliateService.getAddress(
+                affiliateSharePublicKey.toPublicKey(), chaincodeProperties.mainNet
+//                it.toPublicKey(), chaincodeProperties.mainNet
```
rchaing-figure, line 191: meant to remove this?

rchaing-figure (approved):
Good Job!!

p8e-api/src/main/kotlin/io/provenance/engine/event/IndexHandler.kt
```diff
@@ -169,4 +182,34 @@ class IndexHandler(
...
+            affiliateService.getAddress(
+                affiliateSharePublicKey.toPublicKey(), chaincodeProperties.mainNet
+            )
+        }
+        val existingScopeDataAccess = provenanceGrpcService.retrieveScopeData(envelope.data.input.scope.uuid.value).scope.scope.dataAccessList
+        // Only perform job if data access will be updated
+        if (envelopeDataAccess.any { it !in existingScopeDataAccess }) {
+            p8e.Jobs.MsgAddScopeDataAccessRequest.newBuilder()
+                .addAllDataAccess(envelopeDataAccess)
+                .addAllSigners(envelope.data.result.signaturesList.map {
```
scirner22, line 198: By any chance did you test this with a multiparty contract? I'm not positive, but I think the only signer we want is the scope owner. In the case for our scopes that's always a single invoker. If you only ran single party contracts, the `envelope.data.result.signaturesList` should only contain the invoker so it's not a big deal. It would take a multiparty contract to see if this fails or not as is.

scirner22 (approved):
I left one comment about something to try out. Otherwise this looks great!

PR #65, opened 2021-06-30
Keep a consistent signing and verification method for both SmartKey a…

rchaing-figure:
**Context:** This fixes a bug that occurs at the time of bootstrapping for mix affiliates of SmartKey and Database. Before the objects gets pushed to OS, they are signed with the bootstrapping key, in this particular case is an Affiliate with a Database key management type. When executing a contract for a key management type SmartKey, this would fail to retrieve the objects from OS due to signature verification failure, but would retrieve fine with a Database key management type, and cache accordingly, so if the SmartKey affiliate were to run again the next time, it would succeed cause of the cached objects, but if p8e-api were to restart and have that cache cleared that Affiliate with the SmartKey key management type would fail.

**Fix:** Unified the way we signed and verified objects for both SmartKey and Pen class.

### DIFF
--- a/p8e-util/src/main/kotlin/io/p8e/crypto/Pen.kt
+++ b/p8e-util/src/main/kotlin/io/p8e/crypto/Pen.kt
@@ -66,7 +66,8 @@ class Pen(
 
     override fun sign(): ByteArray {
         signature.update(aggregatedData)
-        return signature.sign()
+        // null out the aggregatedData value to reset for next verify/sign
+        return signature.sign().also { aggregatedData = null }
     }
 
     override fun update(data: ByteArray) = signature.update(data)
@@ -85,7 +86,6 @@ class Pen(
                  */
                 if(objSizeIndexer == OBJECT_SIZE_BYTES) {
                     objSizeIndexer = (off + res)
-                    aggregatedData = null
                     aggregatedData = if (aggregatedData == null) {
                         data.copyOfRange(off, off + res)
                     } else {
@@ -104,19 +104,19 @@ class Pen(
     override fun verify(signatureBytes: ByteArray): Boolean {
         signature.update(aggregatedData)
 
-        // Reset the object size indexer.
-        objSizeIndexer = OBJECT_SIZE_BYTES
+        // Reset the object size indexer and null out the aggregatedData value.
+        objSizeIndexer = OBJECT_SIZE_BYTES.also { aggregatedData = null }
 
         return signature.verify(signatureBytes)
     }
 
     override fun initVerify(publicKey: PublicKey) {
-        signature.apply { initVerify(publicKey) }
+        signature.initVerify(publicKey)
         verifying = true
     }
 
     override fun initSign() {
-        signature.apply { initSign(keyPair.private) }
+        signature.initSign(keyPair.private)
         verifying = false
     }
 
--- a/p8e-util/src/main/kotlin/io/p8e/crypto/Pen.kt
+++ b/p8e-util/src/main/kotlin/io/p8e/crypto/Pen.kt
@@ -1,6 +1,9 @@
 package io.p8e.crypto
 
 import com.google.protobuf.Message
+import io.p8e.crypto.SignerImpl.Companion.OBJECT_SIZE_BYTES
+import io.p8e.crypto.SignerImpl.Companion.PROVIDER
+import io.p8e.crypto.SignerImpl.Companion.SIGN_ALGO
 import io.p8e.proto.Common
 import io.p8e.proto.PK
 import io.p8e.proto.ProtoUtil
@@ -16,13 +19,6 @@ class Pen(
     private val keyPair: KeyPair
 ): SignerImpl {
 
-    companion object {
-        // Algo must match Provenance-object-store
-        val SIGN_ALGO = "SHA512withECDSA"
-        val KEY_TYPE = "ECDSA"
-        val PROVIDER = BouncyCastleProvider.PROVIDER_NAME
-    }
-
     val privateKey: PrivateKey = keyPair.private
 
     val signature: Signature = Signature.getInstance(
@@ -33,6 +29,10 @@ class Pen(
     init {
         Security.addProvider(BouncyCastleProvider())
     }
+
+    private var verifying: Boolean = false
+    private var objSizeIndexer: Int = OBJECT_SIZE_BYTES
+    private var aggregatedData: ByteArray? = null
 
     /**
      * Return the signing public key.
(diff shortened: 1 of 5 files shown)


p8e-util/src/main/kotlin/io/p8e/crypto/Pen.kt
```diff
@@ -64,19 +64,61 @@ class Pen(
...
+                    aggregatedData = null
+                    aggregatedData = if (aggregatedData == null) {
+                        data.copyOfRange(off, off + res)
+                    } else {
+                        aggregatedData?.plus(data.copyOfRange(off, off + res))
+                    }
+                } else {
+                    objSizeIndexer += (off + res)
+                    aggregatedData = aggregatedData?.plus(data.copyOfRange(off, off + res))
+                }
```
rchaing-figure, line 97: This is currently used to sign and verify data via the SignatureInputStream for SmartKey, moved this implementation into the Pen class to keep the way of signing/verification consistent between all signer types.

p8e-util/src/main/kotlin/io/p8e/crypto/Pen.kt
```diff
@@ -64,19 +64,61 @@ class Pen(
             .orThrow { IllegalStateException("can't verify signature - public cert may not match private key.") }
     }
 
-    override fun sign(): ByteArray = signature.sign()
-
-    override fun update(data: ByteArray) { signature.update(data) }
+    override fun sign(): ByteArray {
+        signature.update(aggregatedData)
```
rchaing-figure, line 68: Same as the verify, if the signature input stream continues to push data into the update method of the signature instance, the signing/verification cannot properly verify the objects in the below implementation. Keeping in mind that we had to extract the below implementation out of the Signature library so that SmartKey is able to sign the data accordingly under its own Sign api.

p8e-util/src/main/kotlin/io/p8e/crypto/Pen.kt
```diff
@@ -64,19 +64,61 @@ class Pen(
...
+    }
 
-    override fun update(data: ByteArray, off: Int, len: Int) { signature.update(data, off, len) }
+    override fun update(data: ByteArray) = signature.update(data)
+
+    override fun update(data: ByteArray, off: Int, res: Int) {
+        // If off is less then res, these are the data that we care about.
+        if(off < res) {
+            if(!verifying) {
+                val dataSample = data.copyOfRange(off, off+res)
```
rchaing-figure, line 78: This might help reduce the size of the objects we upload to OS, when I was sampling the data with the legacy way of signing (Signature.getInstace() ... ) this included a bunch of trailing insignificant bytes as though it was just empty space. The `off` and the `res` here tells us where the data that we need starts and ends.

p8e-util/src/main/kotlin/io/p8e/crypto/Pen.kt
```diff
@@ -64,19 +64,61 @@ class Pen(
...
-    override fun initVerify(publicKey: PublicKey) { signature.apply { initVerify(publicKey) } }
+        // Reset the object size indexer.
+        objSizeIndexer = OBJECT_SIZE_BYTES
 
-    override fun initSign() { signature.apply { initSign(keyPair.private) } }
+        return signature.verify(signatureBytes)
+    }
+
+    override fun initVerify(publicKey: PublicKey) {
+        signature.apply { initVerify(publicKey) }
```
piercetrey-figure, line 114: I think this could just be `signature.initVerify(publicKey)`, right? The apply seems unnecessary...

p8e-util/src/main/kotlin/io/p8e/crypto/Pen.kt
```diff
@@ -64,19 +64,61 @@ class Pen(
...
+        verifying = true
+    }
+
+    override fun initSign() {
+        signature.apply { initSign(keyPair.private) }
```
piercetrey-figure, line 119: same with the apply here

p8e-util/src/main/kotlin/io/p8e/crypto/Pen.kt
```diff
@@ -64,19 +64,61 @@ class Pen(
...
+            } else {
+                /**
+                 * The downstream (data verification) chunks the data into data size of 8192.
+                 * The data needs to be aggregated to its signing size of 32768 before the
+                 * data can be validated.
+                 */
+                if(objSizeIndexer == OBJECT_SIZE_BYTES) {
+                    objSizeIndexer = (off + res)
+                    aggregatedData = null
+                    aggregatedData = if (aggregatedData == null) {
```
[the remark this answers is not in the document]
rchaing-figure: 🤦

p8e-util/src/main/kotlin/io/p8e/crypto/Pen.kt
rchaing-figure: Yea, I'll clean this up

p8e-util/src/main/kotlin/io/p8e/crypto/Pen.kt
piercetrey-figure: What does `res` stand for? rest? residual? restraint? resign? resplendent? respiratory? resolute?

piercetrey-figure (approved):
a couple of comments/clarifications, but overall if it works then 👍

p8e-util/src/main/kotlin/io/p8e/crypto/Pen.kt
rchaing-figure: I don't know, maybe `result`, just took what was set here: https://github.com/provenance-io/p8e/blob/8ac68c7363d4bcec02c6c147f55e85dc6bc6099f/os-client/src/main/kotlin/io/provenance/os/domain/inputstream/SignatureInputStream.kt#L33 - so i don't confuse the value with `len`

piercetrey-figure: oh, must just be the length of bytes actually read, k

p8e-util/src/main/kotlin/io/p8e/crypto/Pen.kt
scirner22: this class seems like it's very stateful. Does every sign/verify get its own `Pen` class? If it were to be shared across threads it seems like it could have problems signing/verifying two things concurrently.

rchaing-figure: Every affiliate executing something would have to grab their own signer class (Pen in this case), so we might be ok, unless im missing a bigger picture.

PR #67, opened 2021-07-01
Fix for NPE signature

rchaing-figure:
Leaving as a draft as there can be different resolutions.

After discussing with Pierce, it is highly possible that we received an invalid input from the stream. 

As the client side stacktrace shows that there was an attempt to build facts and it has been known that contracts are built with null facts:

```
        at io.p8e.engine.ContractWrapper.toFactInstance(ContractWrapper.kt:132)
	at io.p8e.engine.ContractWrapper.buildFacts(ContractWrapper.kt:105)
	at io.p8e.engine.ContractWrapper.<init>(ContractWrapper.kt:29)
```

Thus receiving NPE at signature update time, when testing with HelloWorld contracts this error didn't come about.

Solutions:
1. Wrap the signature update as in this PR shows and let it error out eventually in the execution lifecycle.
2. Let the client know at this point P8e received invalid/null stuff to sign.

### DIFF
--- a/p8e-util/src/main/kotlin/io/p8e/crypto/Pen.kt
+++ b/p8e-util/src/main/kotlin/io/p8e/crypto/Pen.kt
@@ -65,7 +65,11 @@ class Pen(
     }
 
     override fun sign(): ByteArray {
-        signature.update(aggregatedData)
+        // Only update if the aggregated data is not null
+        if(aggregatedData != null) {
+            signature.update(aggregatedData)
+        }
+
         // null out the aggregatedData value to reset for next verify/sign
         return signature.sign().also { aggregatedData = null }
     }


rchaing-figure:
Also the `signature.update` method moved from the `update` input stream method down stream to the sign method because we needed to manually parse the data needed for the signature in order to support other signer types, thats why this issue happened now and not previously.

piercetrey-figure (approved):
I tried a number of different ways to try and reproduce the original issue (creating contracts with facts that weren't set, functions with fact inputs that weren't set, both single and multi-step versions of these things), and wasn't able to get the same NPE... got other versions of things like expected hash length 64, got 0 or things like that... so, I guess this should at the very least help us figure out what the actual underlying error might be if one crops up

scirner22:
> I tried a number of different ways to try and reproduce the original issue (creating contracts with facts that weren't set, functions with fact inputs that weren't set, both single and multi-step versions of these things), and wasn't able to get the same NPE... got other versions of things like expected hash length 64, got 0 or things like that... so, I guess this should at the very least help us figure out what the actual underlying error might be if one crops up

Did you happen to test the case where you run an update contract that has more than one fact, but only update one of the facts?

rchaing-figure:
> Did you happen to test the case where you run an update contract that has more than one fact, but only update one of the facts?

If I did it right, i don't think I got a NPE out of this (tested off of main)

PR #69, opened 2021-07-07
Npe pen

rchaing-figure:
Make SmartKey signer use the java Signature version of update.

Tested and working :success 

This should make things work the way they were in regards to the Pen class.

### DIFF
--- a/p8e-util/src/main/kotlin/io/p8e/crypto/SmartKeySigner.kt
+++ b/p8e-util/src/main/kotlin/io/p8e/crypto/SmartKeySigner.kt
@@ -5,7 +5,6 @@ import com.fortanix.sdkms.v1.api.SignAndVerifyApi
 import com.fortanix.sdkms.v1.model.DigestAlgorithm
 import com.fortanix.sdkms.v1.model.SignRequest
 import com.google.protobuf.Message
-import io.p8e.crypto.SignerImpl.Companion.OBJECT_SIZE_BYTES
 import io.p8e.crypto.SignerImpl.Companion.PROVIDER
 import io.p8e.crypto.SignerImpl.Companion.SIGN_ALGO
 import io.p8e.proto.Common
@@ -59,9 +58,6 @@ class SmartKeySigner(
     private var verifying: Boolean = false
     private var keyUuid: String? = null
 
-    private var objSizeIndexer: Int = OBJECT_SIZE_BYTES
-    private var aggregatedData: ByteArray? = null
-
     fun instance(keyUuid: String): SmartKeySigner {
         this.keyUuid = keyUuid
         return this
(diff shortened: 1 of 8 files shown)


rchaing-figure:
The other way works as well, but doesn't handles these NPE edge cases and is less abstractive. 

You are probably asking why didn't we do this earlier? I would say a long weekend and a refreshed mind contributed. 😅

scirner22 (commented):
So it looks like the DB based Pen is now pretty much exactly how it looked before all the smart key stuff?

One added comment - can the `SmartKeySigner` take in the signing public key that's stored in the database so that we don't have to call SmartKey `getPublicKey()` in here?

```kotlin
    override fun verify(data: ByteArray, signature: Common.Signature): Boolean {
        val s = Signature.getInstance(signature.algo, signature.provider)
        s.initVerify(getPublicKey())
        s.update(data)
        return s.verify(signature.signature.base64Decode())
    }
```

p8e-util/src/main/kotlin/io/p8e/crypto/SmartKeySigner.kt
```diff
@@ -58,8 +57,12 @@ class SmartKeySigner(
     private var verifying: Boolean = false
     private var keyUuid: String? = null
 
-    fun instance(keyUuid: String): SmartKeySigner {
+    // AffiliateService will pass in a non-null signing public key to instantiate the signer
+    private var publicKey: PublicKey? = null
+
+    fun instance(keyUuid: String, publicKey: PublicKey): SmartKeySigner {
```
piercetrey-figure, line 63: I think @scirner22 brought up the point that this signer is rather stateful and could cause issues with parallel signing/verifying... I think he might be right, I found a few places where we are using a shared signer object in a threadedMap or the like, and I think this class is effectively a singleton since it is the same object that gets returned out of the SignerFactory every time, right? Considering a number of the methods set some private variable containing the current signature being constructed/verified, or update the data being signed over (like some of the update methods), seems like we could end up with a mixture of data and an invalid signature...

Why don't we just move the `keyUuid` and `publicKey` to the constructor, remove this `instance` method and just construct a new SmartKeySigner object in the SignerFactory getSigner method? (I think we would just have to inject the signAndVerifyApi into the SignerFactory constructor, right?

piercetrey-figure: I think even the Pen class could have issues in situations where we have the threadedMap sharing a single `signer` object (i.e. the `loadAllClasses` method of ContractEngine.kt, or `getFactHistory` of IndexGrpc), I would think in these cases, we just need to maybe do multiple signerFactory.getSigner calls within the threadedMap maybe? Idk, let me know if I'm crazy or missing something 😆

p8e-util/src/main/kotlin/io/p8e/crypto/SmartKeySigner.kt
```diff
@@ -84,41 +83,14 @@ class SmartKeySigner(
...
-                    } else {
-                        aggregatedData?.plus(data.copyOfRange(off, off + res))
-                    }
-                } else {
-                    objSizeIndexer += (off + res)
-                    aggregatedData = aggregatedData?.plus(data.copyOfRange(off, off + res))
-                }
-            }
+        if(!verifying) {
+            signatureRequest?.data = data.copyOfRange(off, res)
```
piercetrey-figure, line 87: couldn't this update method be called multiple times in the case of a stream of data... in that case, shouldn't the signatureRequest's data be appended to, otherwise this would just overwrite the data that we will then send to SmartKey for signing, right?

piercetrey-figure (commented):
I guess I have questions 😄

p8e-util/src/main/kotlin/io/p8e/crypto/SmartKeySigner.kt
rchaing-figure: No, for the signing case, the data comes in one big blob to get signed, i have an example of what the data kind of looks like (its huge). It's the verify that comes in smaller broken up chunks, hence the previous appended logic to handle the case.

p8e-util/src/main/kotlin/io/p8e/crypto/SmartKeySigner.kt
rchaing-figure: Thanks for the comments, you brought some good cleanup areas, for `loadAllClasses` and `getFactHistory` the only usage of this was to determine which signer to use to verify as the `update` method were different in both classes, but now that they are the same this probably doesn't need to get pushed all the way down to the `get` of the DefintionService (TBD), but also good point for those threadedMap, if they were both still different it would probably verify against the wrong signer. 

Then for the signing, just did a quick 👀  looks like we are calling for a new signer from the affiliateService right before we sign (`osClient.put()`) unless I'm mistaken we might be safe with any possible mix up. 

I could move the keyUuid and publicKey to a constructor, actually I think I would prefer to have it that way, i recall having it that way but don't remember why I went this `instance` path instead.

piercetrey-figure (approved):
nice, makes me feel less nervous now 😆

PR #73, opened 2021-07-17
Allow signing of empty data for SmartKey

rchaing-figure:
Let SmartKey sign empty data, mirror what the java Signature class currently does, cause there are uses cases where data comes in as empty but still gets signed.


scirner22:
Were you able to replicate the problem we were seeing? I guess it would be a contract where you propose multiple facts and make one of them an empty proto, then hydrate.

scirner22:
Also curious if each singing instance is totally separate, how does one empty proto cause the others in the record group to sign incorrectly?

rchaing-figure:
This might be the case of hash and p8e-contract versions coming from different time spans, with the invalid signature.

Just ran this against SynergyOne on Test with it being SmartKey type, and it executed successfully.

rchaing-figure:
This just allows signing of empty protos to be signed, different from the unable to verify signatures cases that have been floating about.

scirner22 (approved):

PR #76, opened 2021-07-20
Ensure scope separate block

piercetrey-figure:
free specific scopes for further transactions, hopefully a more correct approach than just watching the block height and clearing all scope ids upon an increase in height. This should prevent the same scope from having a contract run on it in different transactions in the same block, which can cause events to remain unprocessed, due to the scope.lastEvent being used to look up envelopes based on execution uuid

### DIFF
--- a/p8e-api/src/main/kotlin/io/provenance/engine/service/ChaincodeInvokeService.kt
+++ b/p8e-api/src/main/kotlin/io/provenance/engine/service/ChaincodeInvokeService.kt
@@ -49,6 +49,14 @@ class ChaincodeInvokeService(
     private val chaincodeProperties: ChaincodeProperties,
     private val provenanceGrpc: ProvenanceGrpcService,
 ) : IChaincodeInvokeService {
+
+    companion object {
+        private val blockScopeIds = HashSet<String>()
+        fun freeScope(scopeUuid: UUID) {
+            blockScopeIds.remove(scopeUuid.toString())
+        }
+    }
+
     private val log = logger()
 
     private val objectMapper = ObjectMapper().configureProvenance()
@@ -75,7 +83,6 @@ class ChaincodeInvokeService(
 
     // non-thread safe data structures that are only used within the worker thread
     private val batch = mutableListOf<ContractRequestWrapper>()
-    private val blockScopeIds = HashSet<String>()
     private val priorityScopeBacklog = HashMap<String, LinkedList<ContractRequestWrapper>>()
     private var currentBlockHeight = 0L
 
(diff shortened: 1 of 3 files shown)


scirner22:
Carry over from slack for github records - we discussed adding a blockheight for when the scope was added to the batch. That way the event stream can iterate over the batch and log at the error level if any of the scopes have been in there for more than N blocks.

p8e-api/src/main/kotlin/io/provenance/engine/service/ChaincodeInvokeService.kt
```diff
@@ -49,6 +49,14 @@ class ChaincodeInvokeService(
     private val chaincodeProperties: ChaincodeProperties,
     private val provenanceGrpc: ProvenanceGrpcService,
 ) : IChaincodeInvokeService {
+
+    companion object {
+        private val blockScopeIds = HashSet<String>()
+        fun freeScope(scopeUuid: UUID) {
```
rchaing-figure, line 55: free in this context the same as release/remove scopes?

piercetrey-figure: yeah, maybe release would be a better word, or unlock 🤔

piercetrey-figure: changed everything to lock/unlock

piercetrey-figure:
@scirner22 I added that in, we will error log a scope uuid (once) if it has been locked for > 10 blocks, let me know what you think

p8e-api/src/main/kotlin/io/provenance/engine/service/ChaincodeInvokeService.kt
```diff
@@ -49,7 +48,37 @@ class ChaincodeInvokeService(
     private val chaincodeProperties: ChaincodeProperties,
     private val provenanceGrpc: ProvenanceGrpcService,
 ) : IChaincodeInvokeService {
-    private val log = logger()
+
+    companion object {
+        private val log = logger()
+
+        private val blockScopeIds = HashSet<String>()
+        private val scopeLockHeights = HashMap<String, Long>()
```
scirner22, line 56: We should probably move these to ConcurrentHashMaps now. Previously only the batching thread was using it, but now it's being unlocked in the event stream thread and reaper as well.

piercetrey-figure: oh good point, I'll update that

piercetrey-figure: updated

scirner22 (approved):

rchaing-figure (approved):

PR #79, opened 2021-07-26
Verify sig decouple

rchaing-figure:
Setup the signature/data verify at the SignatureInputStream level, cleaned up signer variable.

### DIFF
--- a/p8e-util/src/main/kotlin/io/p8e/crypto/SignatureInputStream.kt
+++ b/p8e-util/src/main/kotlin/io/p8e/crypto/SignatureInputStream.kt
@@ -2,23 +2,27 @@ package io.p8e.crypto
 
 import java.io.InputStream
 import java.security.PublicKey
+import java.security.Signature
 
 class SignatureInputStream(
     private val inputStream: InputStream,
-    private val signer: SignerImpl,
+    private val signer: SignerImpl? = null,
+    private val verify: Signature? = null,
     private val signatureBytes: ByteArray? = null
 ): InputStream() {
     override fun read(): Int {
         val res = inputStream.read()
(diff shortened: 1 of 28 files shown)


p8e-util/src/main/kotlin/io/p8e/crypto/SignatureInputStream.kt
```diff
@@ -2,63 +2,67 @@ package io.p8e.crypto
 
 import java.io.InputStream
 import java.security.PublicKey
+import java.security.Signature
 
 class SignatureInputStream(
     private val inputStream: InputStream,
-    private val signer: SignerImpl,
+    private val signer: SignerImpl? = null,
```
piercetrey-figure, line 9: Probably not vital, but since we only ever want either a signer, or a verifier, we could probably make this an enum class with two entries to wrap the signer/verifier and only have one argument and then match based on the type, to make that requirement more clear?

piercetrey-figure (approved):
Just one minor thought for clarification, but overall I get where you are coming from

p8e-util/src/main/kotlin/io/p8e/crypto/SignatureInputStream.kt
rchaing-figure: Yea, i can get this cleaned up, was too excited to get this to TEST and see if it fixes :allthethings 😅

piercetrey-figure: Haha, yeah, don't want to slow that process down 😆

scirner22 (approved):

PR #81, opened 2021-07-30
Create a digest for data stream signing.

rchaing-figure:
Add a message digest in the SmartKey class (output of a hash function) for proper cryptographic signing, adds the necessary padding and out-of-the-box data buffering for streams via `update()`. SmartKey will now sign the hash of the data instead of the data itself (proper way of doing digital signature 🤦 ).

### DIFF
--- a/p8e-util/src/main/kotlin/io/p8e/crypto/SmartKeySigner.kt
+++ b/p8e-util/src/main/kotlin/io/p8e/crypto/SmartKeySigner.kt
@@ -9,7 +9,6 @@ import io.p8e.proto.PK
 import io.p8e.proto.ProtoUtil
 import io.p8e.util.base64Decode
 import io.p8e.util.base64Encode
-import io.p8e.util.base64Sha512
 import io.p8e.util.orThrow
 import io.p8e.util.toPublicKeyProto
 import org.bouncycastle.jce.provider.BouncyCastleProvider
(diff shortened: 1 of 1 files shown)


p8e-util/src/main/kotlin/io/p8e/crypto/SmartKeySigner.kt
```diff
@@ -65,14 +60,19 @@ class SmartKeySigner(
...
 
     override fun update(data: ByteArray, off: Int, res: Int) {
-        signatureRequest?.data = data.copyOfRange(off, res)
+        messageDigest.update(data, off, res)
     }
 
-    override fun sign(): ByteArray = signAndVerifyApi.sign(keyUuid, signatureRequest).signature
+    override fun sign(): ByteArray {
+        // Completes the message digest and adds the necessary padding before signing
+        signatureRequest?.hash = messageDigest.digest()
```
rchaing-figure, line 72: before the standup i had this in the `update` method, so this caused the padding to be off🤦

scirner22: Less bytes over the network as well :)

scirner22 (approved):

piercetrey-figure (approved):
Ohh, yeah, I guess that would do it 😆  Is there any data that needs to be cleaned up in object store as a result of this? Specifically, are there facts for SynergyOne with bad signatures (since they appear to be the only non-test/cronjob affiliate setup with SmartKey)

rchaing-figure:
No, this was only impacting when the bootstrapping key was a SmartKey type, I never converted the bootstrapping key (FigureLending) to SmartKey

PR #83, opened 2021-08-10
Fix for connection reset

rchaing-figure:
Datadog logs were showing some connection reset errors when extension functions were called to convert ephemeral keys to agree keys and transient keys via the SmartKey security object api. The connection reset may have been caused because a new instance of the SecurityObjectApi object was created during this transformation. 

The fix was to add an ExternalKeyCustody class that holds a reference of the security object api that was created during AppConfig time (so that no connection is interrupted).

### DIFF
--- a/p8e-encryption/src/main/kotlin/io/provenance/p8e/encryption/util/ExternalKeyCustodyApi.kt
+++ b/p8e-encryption/src/main/kotlin/io/provenance/p8e/encryption/util/ExternalKeyCustodyApi.kt
@@ -1,0 +1,24 @@
+package io.provenance.p8e.encryption.util
+
+import com.fortanix.sdkms.v1.api.SecurityObjectsApi
+
+/**
+ * A wrapper class for all external key custody apis that are needed.
+ *
+ */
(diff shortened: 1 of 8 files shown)


p8e-encryption/src/main/kotlin/io/provenance/p8e/encryption/util/ExternalKeyCustodyApi.kt
```diff
@@ -0,0 +1,24 @@
+package io.provenance.p8e.encryption.util
+
+import com.fortanix.sdkms.v1.api.SecurityObjectsApi
+
+/**
+ * A wrapper class for all external key custody apis that are needed.
+ *
+ */
+class ExternalKeyCustodyApi(
```
rchaing-figure, line 9: Data class crossed my mind, but thought about future implementation and probably wouldn't make sense, probably need to convert to a factory eventually, but we don't know what the APIs will look like yet.

scirner22 (approved):

piercetrey-figure (approved):

PR #86, opened 2021-08-20
set block height timeout on transactions

piercetrey-figure:
- to prevent transactions from getting stuck in the mempool of a node for whatever reason (usually gas related)
- tested with a positive timeout height of 12 and transactions completed normally, tested with a negative height and transactions were immediately rejected, not sure how to induce a stuck transaction locally

### DIFF
--- a/p8e-api/src/main/kotlin/io/provenance/engine/config/Properties.kt
+++ b/p8e-api/src/main/kotlin/io/provenance/engine/config/Properties.kt
@@ -56,6 +56,7 @@ class ChaincodeProperties {
     @NotNull var txBatchSize: Int = 25
     @NotNull var gasMultiplier: Double = 1.0
     @NotNull var maxGasMultiplierPerDay: Int = 1000
+    @NotNull var blockHeightTimeoutInterval: Int = 12
 }
 
 @ConfigurationProperties(prefix = "elasticsearch")
--- a/p8e-api/src/main/kotlin/io/provenance/engine/config/Properties.kt
+++ b/p8e-api/src/main/kotlin/io/provenance/engine/config/Properties.kt
@@ -56,7 +56,7 @@ class ChaincodeProperties {
     @NotNull var txBatchSize: Int = 25
     @NotNull var gasMultiplier: Double = 1.0
     @NotNull var maxGasMultiplierPerDay: Int = 1000
-    @NotNull var blockHeightTimeoutInterval: Int = 12
+    @NotNull var blockHeightTimeoutInterval: Int = 20
 }
 
 @ConfigurationProperties(prefix = "elasticsearch")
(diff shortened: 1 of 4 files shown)


p8e-api/src/main/kotlin/io/provenance/engine/service/ChaincodeInvokeService.kt
```diff
@@ -209,7 +209,9 @@ class ChaincodeInvokeService(
                 val txBody = batch.map {
                     it.attempts++
                     it.request
-                }.toTxBody()
+                }.toTxBody().toBuilder()
+                    .setTimeoutHeight(currentBlockHeight + chaincodeProperties.blockHeightTimeoutInterval)
```
scirner22, line 213: If https://github.com/provenance-io/p8e/pull/86/files#diff-8c67dfe07e7217aa7039fcc68ae6edb2923d579c1f1d3c63a3daacf49a786664R180 is strung out to the max in the perfect scenario, could currentBlockHeight be at most 5 blocks behind? Not that it's a big deal, but might want to account for that and slightly increase whatever we use for `blockHeightTimeoutInterval`?

p8e-api/src/main/kotlin/io/provenance/engine/config/Properties.kt
```diff
@@ -56,6 +56,7 @@ class ChaincodeProperties {
     @NotNull var txBatchSize: Int = 25
     @NotNull var gasMultiplier: Double = 1.0
     @NotNull var maxGasMultiplierPerDay: Int = 1000
+    @NotNull var blockHeightTimeoutInterval: Int = 12
```
rchaing-figure, line 59: curious why `12` as the magic number?

iramiller:
To test. Set min gas price on one node to a small value. Send tx to that. Note none of the other nodes will accept a tx from that node that is below their minimum gas price.

p8e-api/src/main/kotlin/io/provenance/engine/config/Properties.kt
piercetrey-figure: It is a minute at an approximate 5 second block cut time, but really somewhat arbitrary

p8e-api/src/main/kotlin/io/provenance/engine/service/ChaincodeInvokeService.kt
piercetrey-figure: @scirner22 I'm not quite sure what you are referring to with that link

piercetrey-figure:
@iramiller thanks for the info on how to test that, I tried this and was able to reproduce a transaction eventually timing out due to it getting through the checktx on the node I submitted it to, but then rejected by other nodes.

p8e-api/src/main/kotlin/io/provenance/engine/service/ChaincodeInvokeService.kt
scirner22: It seems like the aim was to have the timeout such that the transaction has 1 minute worth of blocks. I think our notion of currentBlockHeight might be at most 5 blocks behind what the current height really is. If that seems correct, maybe we just want the interval to be 20 blocks or something?

piercetrey-figure: ok, that seems reasonable. I think the current block height should be fairly close usually since we grab it at each loop iteration, but if the batching logic takes a while in some case we could probably fall behind a bit. I'll change the default to 20

scirner22 (approved):

PR #90, opened 2021-10-12
only allow valid event state transitions

piercetrey-figure:
- prevents duplicated events leading to an envelope going from ENVELOPE_RESPONSE back to SCOPE_INDEX for example and the completion event never happening after successful indexing when an affiliate wasn't actively connected

### DIFF
--- a/p8e-api/src/main/kotlin/io/provenance/engine/service/EventService.kt
+++ b/p8e-api/src/main/kotlin/io/provenance/engine/service/EventService.kt
@@ -41,7 +41,8 @@ class EventService() {
 
         val VALID_STATE_TRANSITIONS = mapOf(
             Event.ENVELOPE_FRAGMENT to listOf(Event.ENVELOPE_CHAINCODE, Event.ENVELOPE_ERROR),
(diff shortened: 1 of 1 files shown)


p8e-api/src/main/kotlin/io/provenance/engine/service/EventService.kt
```diff
@@ -38,6 +38,15 @@ class EventService() {
         val SKIPPABLE_EVENTS = listOf(Event.UNRECOGNIZED, Event.SCOPE_INDEX_FRAGMENT)
         val CONNECTED_CLIENT_EVENTS = listOf(Event.ENVELOPE_REQUEST, Event.ENVELOPE_RESPONSE, Event.ENVELOPE_ERROR)
         val UNCONNECTED_CLIENT_EVENTS = Event.values().toList().minus(CONNECTED_CLIENT_EVENTS).minus(SKIPPABLE_EVENTS)
+
+        val VALID_STATE_TRANSITIONS = mapOf(
+            Event.ENVELOPE_FRAGMENT to listOf(Event.ENVELOPE_CHAINCODE, Event.ENVELOPE_ERROR),
+            Event.ENVELOPE_REQUEST to listOf(Event.ENVELOPE_MAILBOX_OUTBOUND, Event.SCOPE_INDEX, Event.SCOPE_INDEX_FRAGMENT),
```
scirner22, line 44: `REQUEST` doesn't go to `CHAINCODE`?

PR #95, opened 2021-11-17
Add connection timeout to streaming event call

piercetrey-figure:
add interceptor to implement a connection timeout when connecting to the streaming event call

- Can't use a regular grpc deadline, since that can't be per-message and messages can be sent far into the future when the stream is open for a long time

Hopefully this fixes the occasional mysterious hang that locks up everything for client watchers


scirner22 (approved):
This looks promising

rchaing-figure (approved):

kherzinger-figure (approved):

PR #96, opened 2021-11-22
Query chain directly when fetching latest scope for IndexGrpc

piercetrey-figure:
- Allows scopes that were created by new scope sdk to be queried properly via the old p8e
- This is an important piece of the los transition to the new sdk (since the final AssignToServicer contract requires scope sdk multiparty support to move that to the new sdk)
- Also prevent error resulting from incoming events that have no `lastEvent.executionUuid` set (i.e. scopes written by something other than the legacy p8e system, such as the new scope sdk, p8e doesn't need to do anything with these scopes anyways)
- Note: I went this route for fixing the query issue, rather than allowing incoming events w/o a last execution uuid from being inserted into the index_scope table and leaving the query alone, as the new scope sdk will write piecemeal record updates for existing scopes, so a ScopeUpdated event won't necessarily be emitted, and therefore subsequent updates to the scope wouldn't be recorded in the p8e index_scope cache, so it seems better to just use the chain as the source of truth...
- Note: This won't return the block number/block transaction index for these types of scopes since that information doesn't exist on the scope response from chain

### DIFF
--- a/p8e-api/src/main/kotlin/io/provenance/engine/grpc/v1/IndexGrpc.kt
+++ b/p8e-api/src/main/kotlin/io/provenance/engine/grpc/v1/IndexGrpc.kt
@@ -159,13 +159,13 @@ class IndexGrpc(
                 .map { it.toUuidProv() }
                 .filterNot { fetchedUuids.contains(it) }
                 .also { log.debug("Attempting to fetch an additional ${it.count()} scopes from chain") }
-                .map {
+                .mapNotNull {
                     try {
                         provenanceGrpcService.retrieveScope(it)
                     } catch (e: Exception) {
                         null
                     }
-                }.filterNotNull()
+                }
                 .contractScopesToScopeWrappers()
 
             log.debug("Fetched additional ${chainScopes.scopesCount} scopes from chain")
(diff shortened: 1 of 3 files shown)


p8e-api/src/main/kotlin/io/provenance/engine/grpc/v1/IndexGrpc.kt
```diff
@@ -132,9 +144,38 @@ class IndexGrpc(
...
+
+        if (indexedScopes.scopesCount != request.uuidsCount) {
+            val fetchedUuids = indexedScopes.scopesList.map { it.scope.uuid.toUuidProv() }.toSet()
+
+
+            val chainScopes = request.uuidsList
+                .map { it.toUuidProv() }
+                .filterNot { fetchedUuids.contains(it) }
+                .also { log.debug("Attempting to fetch an additional ${it.count()} scopes from chain") }
+                .map {
```
scirner22, line 162: is this equivalent to `mapNotNull`?

scirner22 (approved):

p8e-api/src/main/kotlin/io/provenance/engine/grpc/v1/IndexGrpc.kt
piercetrey-figure: oh yeah, I think so, good call I'll change that

PR #100, opened 2021-12-03
send an immediate event on client watch connect to force grpc headers through

piercetrey-figure:
- prevents ConnectionTimeoutInterceptor from timing out when an envelope isn't executed within the timeout window
- can only send this event when the client supports it (to avoid the watch dying due to 'unknown event type'), so I had to add some stuff in to send the p8e-sdk version as a GRPC header (could have sent an ENVELOPE_ACCEPTED event, but this seemed better)

### DIFF
--- a/p8e-api/src/main/kotlin/io/provenance/engine/grpc/observers/EnvelopeEventObserver.kt
+++ b/p8e-api/src/main/kotlin/io/provenance/engine/grpc/observers/EnvelopeEventObserver.kt
@@ -15,6 +15,7 @@ import io.p8e.proto.Envelope.EnvelopeEvent.Action.*
 import io.p8e.proto.Envelope.EnvelopeEvent.EventType
 import io.p8e.proto.PK
 import io.p8e.util.*
+import io.provenance.engine.const.P8eVersions
 import io.provenance.p8e.shared.extension.logger
 import io.provenance.engine.domain.AffiliateConnectionRecord
 import io.provenance.engine.domain.ConnectionStatus.CONNECTED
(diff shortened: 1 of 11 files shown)


p8e-api/src/main/kotlin/io/provenance/engine/grpc/observers/EnvelopeEventObserver.kt
```diff
@@ -136,7 +138,31 @@ class EnvelopeEventObserver(
...
+                    .let(queuer::queue)
+            } else {
+                logger().info("Skipping ENVELOPE_WATCH_CONNECTED event for unsupported client (version ${clientVersion()})")
+            }
+        }
+    }
+
+    private fun clientSupportsEnvelopeWatchConnected(): Boolean = try {
+        SemVer.parse(clientVersion()).let {
+            it.compareTo(SemVer.parse("0.8.22")) > 0
```
scirner22, line 161: Maybe put this in a const with a comment to the commit where it was added?

p8e-sdk/src/main/resources/version.properties
```diff
@@ -0,0 +1 @@
+version=${version}
```
scirner22, line 1: Does the semver plugin manage setting this on builds or something?

piercetrey-figure: this block in the gradle file populates the version number https://github.com/provenance-io/p8e/pull/100/files#diff-49a96e7eea8a94af862798a45174e6ac43eb4f8b4bd40759b5da63ba31ec3ef7R67

I was hoping there was some more automatic way to pull the version in, but apparently not

piercetrey-figure: the semver plugin is just to assist in comparing semver strings to determine if it is a newer or older version

p8e-api/src/main/kotlin/io/provenance/engine/grpc/observers/EnvelopeEventObserver.kt
piercetrey-figure: Good call, added that

p8e-sdk/src/main/resources/version.properties
scirner22: Ah got it.

scirner22 (approved):

PR #102, opened 2021-12-09
look up signing key via AffiliateService when erroring envelope

piercetrey-figure:
The ol' different signing/encryption keys strikes again!

Homepoint originator in prod has different signing/encryption/auth keys, and when the servicing affiliate rejected an envelope, the error failed to propagate back to the Homepoint originator, since the incoming mail public key is the encryption key, while the envelope table key is the signing key. This ensures the correct key is used when processing incoming mail

- ensures correct key is used for envelope lookup, preventing error from being dropped


scirner22 (approved):

rchaing-figure (approved):
:shakesfist

PR #107, opened 2022-03-01
configurable options for fetching/injecting gas price dynamically

piercetrey-figure:
* Configuration properties to determine where to fetch gas price from, or what gas price to use (/as a fallback)
* Figure provides a gas price at https://test.figure.tech/figure-gas-price
   * Will be a prod version as well soon, can be any url that returns json like below (gasPriceDenom _has_ to be `nhash` for now, or will be rejected and default will be used)
```
 {
    "gasPrice": 19050,
    "gasPriceDenom": "nhash"
}
```

### DIFF
--- a/p8e-api/src/main/kotlin/io/provenance/engine/service/GasPriceService.kt
+++ b/p8e-api/src/main/kotlin/io/provenance/engine/service/GasPriceService.kt
@@ -1,0 +1,57 @@
+package io.provenance.engine.service
+
+import feign.RequestLine
+import io.provenance.p8e.shared.extension.logger
+import org.springframework.stereotype.Component
+import java.lang.IllegalStateException
+import java.time.Duration
+import java.time.OffsetDateTime
+
+interface GasPriceService {
+    fun getGasPriceNHash(): Double
+}
+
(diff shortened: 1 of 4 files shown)


p8e-api/src/main/kotlin/io/provenance/engine/service/GasPriceService.kt
```diff
@@ -0,0 +1,57 @@
...
+    }
+}
+
+class ConstantGasPriceService(val gasPrice: Double): GasPriceService {
+    override fun getGasPriceNHash(): Double = gasPrice
+}
+
+class CachedGasPriceService(val inner: GasPriceService, val cacheDuration: Duration): GasPriceService {
+    private var lastFetched: OffsetDateTime = OffsetDateTime.now()
+    private var lastFetchedPrice: Double = inner.getGasPriceNHash()
```
rchaing-figure, line 39: Might be cleaner to have these variables in a data class? if it's too much trouble, it's fine.

rchaing-figure (approved):

p8e-api/src/main/kotlin/io/provenance/engine/service/GasPriceService.kt
piercetrey-figure: I do like that idea, just pushed something up

PR #108, opened 2022-04-27
configurable batch size for contract/scope spec txs

piercetrey-figure:
* also set a timeout height on these txs to prevent them from getting 'stuck'


rchaing-figure (approved):

afremuth-figure (commented):
👍

